PR #91, opened 2022-03-10
Fix issue where ORCID login was not working because of null values

nathan-at-jisc:
The purpose of this PR was to fix the issue that was causing certain users to not be able to login. This was caused by possible null values for `{ day: null }`.

This was fixed with more optional chaining.

Also fixed the issue where certain publications were not displaying. This was an easy fix, i needed to display if either `linkedTo` or `linkedFrom` have length, previously I was only checking `linkedTo`.


shaunajisc:
lgtm! 😁

shaunajisc (dismissed):

shaunajisc (approved):

PR #102, opened 2022-03-25
Implementing OpenSearch (OCT 183)

nathan-at-jisc:
The purpose of this PR was to implement OpenSearch. As you can imagine, this involves quite a few things:

- Using the prisma seed data to also push `LIVE` data to OpenSearch.
- Add OpenSearch to the docker-compose file.
- Create an OpenSearch TF module.
- Update env variables to use SSM parameters (user, password and endpoint).
- When adding data to OpenSearch, using a library called `htmlToText` to strip out html from `content` and create a field called `cleanContent`. This is the field that should be searched against.
- Removed `orderBy` and `orderDirection` from publications. This was because it wasn't working quite right as OpenSearch should be first ordering by its `_score`. We'll investigate this in another PR.
- Most file changes are changing `prisma.model` to `client.prisma.model`, as we now also have `client.search`. 

The way we are now using OpenSearch is as follows:
1. When a publication changes status to `LIVE`, we create a record using the id from postgres. If that id is already in use, it will update it.
2. When a user hits `GET /publications`, it is only searching `LIVE` publications.
3. We first search `OpenSearch` to retrieve the pagination data, along with a list of ids for publications.
4. We use this array of ids to pull the data from `prisma` so that we can keep relations in tact (as we are not storing links, status history in open search).

This may seem wasteful, and it kinda is. However, it doesn't add that much time to each request (probably around 50-100ms). It means that we don't have to keep all relations in OpenSearch.

---

### Tests:

At this stage, there are no tests.


AshleyRedmanJisc (approved):
Browse page issue with fetching publications, but the rest to do with opensearch looks good. Think we need to fix any found issues separately, so happy to approve now and fix in a later pr.

PR #103, opened 2022-03-25
Co-authors (OCT-43)

CarlyRich:
The purpose of this PR was...

---

### Acceptance Criteria:
- A publication cannot be published unless all the co-authors have validated “positively”.

- At any point during the publication's draft state, co-authors can be added/removed.

- Co-authors validate by clicking ‘yes’ in an email, which will direct to the Octopus site, where they will login (for the first time, or otherwise). They will be able to view the draft publication in a read only state. 

- They can then leave on “no” if an edit is required or change to ‘yes’ when they have read through and are happy with the draft publication. These co-authors will all have Octopus accounts.

- When co-authors select any of the options, the submitter will be notified.

- If a co-author validates “no” on the email, they will automatically be removed from the list in the publication.

- There is a 30 day time limit for validation. After this the submitter will need to resend. 

- “Remind all” button to nudge co-authors to validate. Individual co-authors can also be nudged. 

- Email is not customisable, for now.

API endpoints created:

create = create a coauthor (used by author)
resend = resend a coauthor email (coauthor gets a new code) (used by author)
delete = delete a coauthor from the table (used by author)
reset = resets all coauthors codes and coauthors confirmation to false (used by author)

confirm = confirm that they are a coauthor and attach a user (used by coauthor)
deny = deny that they are a coauthor and delete record from table  (used by coauthor)
update = update the confirmedCoAuthor field to true or false depending on if an edit is needed or not (used by coauthor)

API endpoints changed:
Endpoint for getPublications, getUser and getUsers now all return coAuthors.
Endpoint for getPublications logic now shows draft publications in readonly state for co-authors.

todo// 
code expires after 30 days
finish update endpoint after changes in meeting
change publish endpoint to add a check that all coauthors are returning true in confirmedcoauthor.

---

### Tests:

---


nathan-at-jisc:
Added to this PR, there are now 4 endpoints.

## Create co-author

This endpoint can be "hit" by the publication creator. It does not accept duplicate email addresses.

> POST /publications/{id}/coauthor
```json
{
  "email": "nathan.sainsbury@jisc.ac.uk"
}
```

## Delete co-author

This endpoint can be "hit" by the publication creator. It removes the co-author from the publication.

> DELETE /publications/{id}/coauthor/{coauthor}

## Link co-author

This endpoint is used to link a user to a co-author. It also allows a user to "un-link" the co-author. It has a parameter, `approve`. If `approve` is `true`, then the user making the request will be linked to the co-author (details in body). If `approve` is `false`, authentication is not needed and the co-author will be deleted.

> PATCH /publications/{id}/link-coauthor

```json
{
  "email": "nathan.sainsbury@jisc.ac.uk",
  "code": "code_goes_here",
  "approve": true
}

## Update co-author status

This endpoint is used by authenticatd co-authors. Users can `confirm` whether they agree the publication is ready to go live.

> PATCH /publications/{id}/coauthor

```json
{
  "confirm": true
}
```

api/src/components/coauthor/controller.ts
```diff
@@ -0,0 +1,204 @@
...
+
+        // Is this user the author of the publication?
+        if (publication?.user.id !== event?.user.id) {
+            return response.json(403, {
+                message: 'You do not have the right permissions for this action.'
+            });
+        }
+
+        // Is the publication in draft?
+        if (publication.currentStatus === 'LIVE') {
```
AshleyRedmanJisc, line 27: Because of the hidden status, maybe we check that it is only draft and not if it is live, i.e
```
if (publication.currenttatus !== 'DRAFT')
```

So publications with hidden status dont pass this check too.

api/src/components/coauthor/controller.ts
```diff
@@ -0,0 +1,204 @@
...
+
+        // Is this user the author of the publication?
+        if (publication.user.id !== event.user.id) {
+            return response.json(403, {
+                message: 'You do not have the right permissions for this action.'
+            });
+        }
+
+        // Is the publication in draft?
+        if (publication.currentStatus === 'LIVE') {
```
AshleyRedmanJisc, line 72: Same as the above, as this logic is duped, maybe a middleware or similar can be setup to run these two checks

AshleyRedmanJisc (commented):
The code looks fine to me and follows the mental flow of the feature. Just to confirm the one comment i made & that emails arn't yet setup for this?

shaunajisc (approved):
LGTM :)

api/src/components/coauthor/controller.ts
nathan-at-jisc: Good point, I think we should leave til we have `hidden` status.

api/src/components/coauthor/controller.ts
nathan-at-jisc: Same as above

PR #105, opened 2022-03-28
Exclude ids from GET /publications

nathan-at-jisc:
The purpose of this PR was to modify the `GET /publications` to include an `exclude` query parameter.

> GET /publications?exclude=id1,id2

This endpoint is primarily used for finding a list of "linkable" publications. We will pass in a comma seperated list of ids that should not be returned from the DB.


AshleyRedmanJisc (approved):
gg

PR #106, opened 2022-03-31
Creating an API endpoint to create ratings for publications

shaunajisc:
The purpose of this PR is to create an API endpoint so authenticated users can create ratings out of 10 for publications based on three rating categories, which are specific to the type of publication they are rating.

The route and body for requests will be structured like this: 

**Route:** 
`POST /publications/{id}/ratings`

**Body:** 
`{
    "type": "PROBLEM_IMPORTANT",
    "value": "8"
}`

and we will expect a **valid response** that looks like this: 
```
{
    "id": "publication-problem-live-test-user-2-PROBLEM_IMPORTANT",
    "publicationId": "publication-problem-live",
    "userId": "test-user-2",
    "rating": 8,
    "category": "PROBLEM_IMPORTANT"
}
```
**The categories are specific to the publication type and they are as follows:** 
image

**and the schema for how the data will be stored looks like this:** 
image

### Acceptance Criteria:
- User must be logged in to submit a review 
- User cannot submit more than one review per publication per publication type 
- User can’t leave a rating on their own publication 
- A user may submit a rating for each category individually i.e. they could rate a publication on originality on Tuesday, then rate it for relevance on Thursday. 
- Each publication can only have three rating categories 
- The rating categories can have different names depending on the publication type but their names cannot be changed 
- Rating types that are displayed as the same, i.e Hypothesis and Methods both have a rating type of “concise”, in the backend these are still stored separately
- Ratings can be amended/ overwritten once submitted 
- Ratings will not use stars to indicate quality
- Ratings do not persist between versions 

### Tests:
- The user can leave a review on another user’s publication 
- The user cannot leave a review on their own publication 
- The user needs to be authenticated to leave a review 
- The publication needs to exist for user to leave a review 
- The rating value cannot be less than 0 or more than 10 
- The user must submit a valid rating category based on the publication type
- If the publication is in draft state we cannot leave a review
- If the user leaves a review and then changes it, the database only stores the latest


nathan-at-jisc (approved):
LGTM

PR #111, opened 2022-04-11
Ratings UI (OCT-189)

AshleyRedmanJisc:
NOTE: **The major of 'files changed' includes the local font files.**

The purpose of this PR was to create the UI for creating & display ratings on the publication single view page.

This PR covers [Jira ticket 189](https://jiscdev.atlassian.net/jira/software/projects/OCT/boards/836?selectedIssue=OCT-189).

This PR has also fixed the Google font issues we were facing when deploying to production, i.e Google font CDN not correctly pulling in the font faces. To handle this, this PR includes all the fonts locally & referenced locally so no more CDN calls. 

This PR also introduces the concept of a feedback notification, i.e when an action occurs we can not push the response of that action to an indicating notification card which will be helpful to users know when something has completed or failed. In this case, to help the user know when they rating has been successfully saved. 

---

### Acceptance Criteria:

A user should be able to see the current ratings of a publication when viewing a publication.A user should be able to select ‘Rate this publication’ and see a model where they can rating the current publication. The rating ‘types’ will be based on what the publication type is. All that logic is handled via the API. 

The user will not be able to see the total aggregate rating for the publication (mean of the three categories). 

We may look at options for viewing the full list of ratings that have been made, but these should be anonymous on the publication page (ie it users name should not appear next to their rating). This ticket may cover adding rating values within search results, but I think that is still on going work.

---
### Tests: As of 14/04/2022 (12 mid day)

Screenshot 2022-04-14 at 12 03 20

---

### Screenshots:

https://user-images.githubusercontent.com/81176162/163196239-dce0779e-565c-43a9-a84f-fbc7d528f678.mov


shaunajisc (dismissed):
Lgtm :)

ui/src/lib/interfaces.ts
```diff
@@ -153,3 +154,54 @@ export interface LicenceTypeShape {
     description: string;
     link: string;
 }
+
+export interface Rating {
+    aggregate: RatingEntry[];
+    overall: {
+        _avg: {
+            rating: number;
```
nathan-at-jisc, line 162: `rating: number | null`

AshleyRedmanJisc: Applied in last commit

nathan-at-jisc (approved):
LGTM

PR #115, opened 2022-04-19
Adding/deleting images via the API (OCT-196)

nathan-at-jisc:
The purpose of this PR was to create the ability to upload and delete images via the API. A lot went into this PR:
- Updated Terraform to include an `s3` module, that will create the `s3` bucket needed to store images per environment (i.e `int`, `prod`, etc).
- Updated `docker-compose` to add `localstack`, enabling us to run a copy of `s3` locally.
- Updated seed command so that it checks if a local `s3` bucket exists, if it doesn't, it will create, `science-octopus-publishing-images-local`.
- Added `image` model to prisma. This does not store the image, only a reference to it (i.e `id`, `name`, `user` and `createdAt`).
- Added helper function to UI, to get Base64 of image.

> POST /images
```json
{
  "name": "my-first-image",
  "image": "data:image/jpeg;base64,/9j/4AAQSkZJRgABAQAAAQABAAD/2wCEAAUFBQUFBQYGBgYICQgJCAwLCgoLDBINDg0ODRIbERQRERQRGxgdGBYYHRgrIh4eIisyKigqMjw2NjxMSExkZIYBBQUFBQUFBgYGBggJCAkIDAsKCgsMEg0ODQ4NEhsRFBERFBEbGB0YFhgdGCsiHh4iKzIqKCoyPDY2PExITGRkhv..."
}
```

> DELETE /images/{id}

> GET /images
---

### Acceptance Criteria:
- Only authenticated users can upload images.
- Only `jpg`, `jpeg` and `png` can be uploaded.
- Only the owner of the image can delete their image.
- When retrieving a list of images, only you can see your list.
- Anyone can view any image.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,8 @@
 node_modules
 jspm_packages
 
+.localstack
+
 # Serverless directories
 .serverless
 
--- a/api/package-lock.json
+++ b/api/package-lock.json
@@ -1323,6 +1323,11 @@
                 "defer-to-connect": "^1.0.1"
             }
         },
+        "@tokenizer/token": {
+            "version": "0.3.0",
+            "resolved": "https://registry.npmjs.org/@tokenizer/token/-/token-0.3.0.tgz",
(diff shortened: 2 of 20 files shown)


AshleyRedmanJisc (approved):
LGTM. TF planned and infra up from changes.

PR #122, opened 2022-04-27
Fixed all linting & formatting errors.

AshleyRedmanJisc:
The purpose of this PR was to fix any linting & formatting errors found after adding the a11y ESlint plugin. The a11y plugin takes president over the supplied Next ESlint rules which is good as a11y is high priority.

After addressing all the errors, there are now no disabled rules for a11y. 

Screenshot 2022-04-27 at 14 11 02
Screenshot 2022-04-27 at 14 10 34

In terms of the API, certain rules have now been disabled with the aim to reenable over time. 

Prior to this PR, ESLint was _not_ working in the API due to a file error, so fixing that revealed an array of warnings & errors. Some minor ones have been addressed, but it's more so determining what rules we want to turn off, set to warn and set to error. If an ESLint error is found, this _will_ fail the GH action that checks the linting (when that is setup for the api, currently only the UI due to the difficult nature of running api tests in GH)


nathan-at-jisc (approved):
LGTM

PR #126, opened 2022-04-29
Create and delete a bookmark API endpoints

shaunajisc:
The purpose of this PR was to create two endpoints which allow authenticated users to create a bookmark for a publication and another to delete a bookmark.

A user cannot bookmark their own publication, or a publication that they have co-authored. They can only delete a bookmark that they have created. 

This PR includes changes to the schema, the tests for both endpoints, new seed data, an updated Serverless file and the creation of the endpoints (controller/ route) etc. 

I have created two different endpoints that look like this (neither of the endpoints take a body): 

- One to create the bookmark:
**Path**: `POST /publication/{publicationId}/bookmark`
Which receives a successful response that looks like: 
```json
{
    "message": "You have bookmarked this publication."
}
```

- Another to delete the record of the bookmark: 
**Path** `DELETE  /publication/{publicationId}/bookmark`
Which receives a successful response that looks like: 
```json
{
    "message": "Bookmark deleted from this publication."
}
```

### Acceptance Criteria:

- Users must be authenticated to both create and delete a bookmark 
- A user can only create one record of a bookmark per publication 
- Only the creator of the bookmark can delete the bookmark 
- Users cannot bookmark publications they have authored or co-authored 
- Users can only bookmark publications in a `'LIVE'` state

### Tests:
```
 PASS  src/components/bookmark/__tests__/createBookmark.test.ts (9.306 s)
  create a bookmark
    ✓ Create a bookmark (833 ms)
    ✓ Cannot create a bookmark as an un-authenticated user (482 ms)
    ✓ Cannot create a bookmark if the user is the author of the publication (589 ms)
    ✓ Cannot create a bookmark on a publication that is not live (536 ms)
    ✓ Cannot create two bookmarks on the same publication as one user (532 ms)
    ✓ Cannot create bookmarks a publication that does not exist (544 ms)
```

```
 PASS  src/components/bookmark/__tests__/removeBookmark.test.ts (8.792 s)
  Remove a bookmark
    ✓ Delete a bookmark (840 ms)
    ✓ Cannot delete a bookmark as an un-authenticated user (505 ms)
    ✓ Cannot delete a bookmark that another user has created (512 ms)
    ✓ Cannot delete a bookmark that does not exist (562 ms)
```


AshleyRedmanJisc:
@shaunajisc I think main needs to be pulled because of the found conflict with the schema file

nathan-at-jisc (dismissed):
LGTM

PR #137, opened 2022-05-18
OCT-209 Supply and verify email address

tomgiddingsjisc:
The purpose of this PR was to provide a means to allow an email address to be gathered (and verified) from authenticated users.

#### API
- **env file changes** to add `VALIDATION_CODE_EXPIRY` and `VALIDATION_CODE_ATTEMPTS` variables.
- **DB migration** to add Verification table.
- **New endpoints** for requesting a verification email with code, and to verify that an entered code is correct.
- **User endpoint bugfix** to prevent private fields (such as email) from being accessible to public API users (restricts to account owner via token).
- **Added `updateEmail` function** to user component.
- **API tests** for verification endpoints.
- **Email template** for verification.

#### UI
- **Verify page** with copy changes for new users and existing users wishing to update their email.
- **Extended button component** to allow more control over text size and padding.
- **Publication page** will prevent actions being carried out if a user does not have an email address on their account.
- **Account/My Publications page** now displays a user email and provides an option to update their address if required.
- **Banner** shown to registered users who have not yet verified their email address (see screenshots).

---

### Acceptance Criteria:

As per [ticket](https://jiscdev.atlassian.net/jira/software/projects/OCT/boards/836?selectedIssue=OCT-209):

> - A new authenticated user **should** be prompted to supply their email address on logging in.
> - As part of the prompt, an authenticated user **should** see the public information provided by ORCID (such as name and affiliation). This should not be editable.
> - An email containing a unique (short) code **should** be sent to the user.
> - A unique code **should** be usable only once to verify the associated user.
> - A unique code **should** expire on three failed attempts and return the user to re-enter their email address.
> - A unique code **should** expire after 10 minutes.
> - In order to verify their email address, a user **should** enter the corresponding unique code at the prompt.
> - Entering an incorrect code will **should not** verify an email address.
> - An option to send a new code (and email) **should** be present at the code entry step.
> - An option to return to the previous step (to re-enter email) **should** be present at the code entry step.
> - An email address for new or existing users **should** only be used when it has been verified.
> - Until an email address is verified, a new authenticated user **should not** be authorised to access any area of the site which requires them to log in.
> - Once an email address is verified, an authenticated user **should not** be prompted to verify their email address on subsequent login attempts.
> - If a new authenticated user has left the site without verifying their email address, they **should** be returned to the prompt requiring them to enter their email address.
> - An existing authenticated user **should** be able to update and re-verify their email address using the same method as new users.
> - On providing an updated email address, an existing authenticated user **should** still be able to access all areas of the site, regardless of the new email verification state.
> - Until a new, updated email address is verified, the existing email address **should** remain in use and verified.

---

### Tests:

Added API tests for new `verification` component.

---

#### Verify prompt

localhost_3001_verify_newUser=true (2)
localhost_3001_verify_newUser=true (4)

localhost_3001_verify_newUser=true (1)
localhost_3001_verify_newUser=true (3)

#### Video

https://user-images.githubusercontent.com/86601264/169107070-0aaeacec-88df-4a55-864b-788c6abc27ae.mov

#### Banner

localhost_3001_browse (2)
localhost_3001_browse (1)


ui/src/components/Header/index.tsx
```diff
@@ -47,6 +49,20 @@ const Header: React.FC<Props> = (props): React.ReactElement => {
                         </div>
                     </div>
                 </div>
+                {/* Confirm email banner */}
+                {user && !user?.email && window.location.pathname !== Config.urls.verify.path && (
```
AshleyRedmanJisc, line 53: I'd be tempted to have this in its own comp added added in the layouts rather directly in the head, but thats a preference

ui/src/pages/verify.tsx
```diff
@@ -0,0 +1,305 @@
...
+            // verify the supplied code, and if successful,
+            const getToken = await api.post<{ token: string }>(
+                `${Config.endpoints.verification}/${user?.orcid}`,
+                { code },
+                user?.token
+            );
+
+            // If success, decode the JWT, set the updated user, and redirect to state
+            if (getToken.status == 200) {
+                const decodedJWT = Helpers.setAndReturnJWT(getToken.data.token) as Types.UserType;
```
AshleyRedmanJisc, line 103: Around this time from the video there is a visual bug where user is re added to the state and because user object X is different from user object Y at this point, so the nav is doubling up. React is rendering them both before routing to the new page where the state is normalised again. So i would guess to get around this, we would log the user out, &. re log them in again, and that process would re get the user

AshleyRedmanJisc (commented):
Made a few comments, the page its self has an interesting interaction that I think we can look at

api/src/components/verification/controller.ts
```diff
@@ -0,0 +1,80 @@
...
+) => {
+    try {
+        const verification = await verificationService.find(event.pathParameters.id);
+
+        if (!verification) {
+            return response.json(404, { message: 'Not found' });
+        }
+
+        // If code has expired, remove it from the database and return appropriate error
+        if (moment().diff(verification.createdAt, 'minutes') > Number(process.env.VALIDATION_CODE_EXPIRY)) {
```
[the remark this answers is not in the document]
tomgiddingsjisc: It's already present and used elsewhere, but will swap over to luxon with a view to standardising!

ui/src/components/Header/index.tsx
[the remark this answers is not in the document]
tomgiddingsjisc: It does work, but good shout on using Router!

tomgiddingsjisc: I think if there were potential for it to be used anywhere else I would, but it feels really unlikely?

ui/src/pages/verify.tsx
tomgiddingsjisc: Sure - I'll have a look at this. Not sure if logging out/in will be too jarring but definitely needs a fix :)

ui/src/components/Header/index.tsx
AshleyRedmanJisc: 😄  comes down the preference thing again i guess. I see components as just pieces of UI, not so much concerned of their reuse. Like if i was putting the header component in isolation in storybook or something, is that other bit of UI tired to the header, I dont know 🤷🏼

tomgiddingsjisc: As is, it's designed to only be shown in the header given the conditions. Given its minimal size and importance, I think it's probably tied to the header in this case.

I'd be inclined to keep this one as is - though I can imagine the complexity increasing in future if we want to keep using that space for other things!

ui/src/pages/verify.tsx
```diff
@@ -0,0 +1,305 @@
+import Head from 'next/head';
```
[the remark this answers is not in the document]
tomgiddingsjisc: That should be re-jigged!

ui/src/pages/verify.tsx
tomgiddingsjisc: Found this alternative fix:

https://github.com/JiscSD/octopus/pull/137/commits/8d2288ddcfc8cc2eb2d9cf0a07e0b1495a7015c6

Seems to do the trick, what do you think?

AshleyRedmanJisc: 👍🏼  bit of an odd one, but should be fine to be honest, especially if there is a navigation change.

Issue #151, opened 2022-05-20
Date selector needs to be disabled when author search is selected, causing run time error.


AshleyRedmanJisc:
https://user-images.githubusercontent.com/81176162/169509969-1870b592-c907-4d0e-bc79-e9676d7b04b3.mov

PR #168, opened 2022-06-23
OCT-issue-151

reubenporterjisc:
Hides the date pickers on the `/search` page when the chosen search type is 'users'. This fixes a bug where a runtime error would occur as a date filter is only valid for publications.

I also spotted the empty results alert contained placeholder text so have populated that with some more useful text.

Issue: #151 

---

### Acceptance Criteria:

Go to the search page and change the search type to users, the date pickers should disappear. Changing back to publications should make them reappear again.


shaunajisc (approved):
LGTM :)

PR #182, opened 2022-07-06
Allow users to add an 'Affiliated Organisations' and/or 'Funders' statement if no organisation has been added - UI

shaunajisc:
The purpose of this PR was to remove the 'Affiliated Organisations' and 'Funders' limit which only shows free text field when organisation has been added (within the publication creation process). Initially you could only add a statement if you had added at least one organisation, but now you can at a statement regardless. This PR contains amendments to the ROR form and amendments to the publication index page with the UI.

---

### Screenshots:
Screenshot 2022-07-06 at 15 01 48
Screenshot 2022-07-06 at 15 01 36


MattMcGowanJisc (approved):
LGTM!

Issue #179, opened 2022-07-05
Site design adaptability for smaller screens 


lharre:
The site design is not suiting smaller laptop/tablet sizes. This issue was noted on a MacBook (Retina, 12-inch). 

Specific issues:

- Column width on publication page is adapting for mobile as standard, (so nav bar appears at top etc) not ideal UX. Can we make the break point more generous?

-  The search magnifying glass icon on the Search page covers the text

![image](https://user-images.githubusercontent.com/97461921/177306820-dcc3dda0-342c-48c4-b66d-e65dc61ff329.png)

PR #185, opened 2022-07-06
Amend the design of the search bar within the 'Search' page to make it more usable for small screens

shaunajisc:
The purpose of this PR was to amend the length of the search bar to make it more adaptable for smaller screens - as per [this Github issue ](https://github.com/JiscSD/octopus/issues/179). 

I have amended the search bar so that it takes up the full width left of the row for lg + sized screens. Looking at the seed data, it seems that the titles of publications can get quite lengthy so I can assume that a user's search query could also get quite lengthy. So it makes sense to allow the search bar to take up the maximum width to allow users to type as much as they can without being cut off, whilst also not compromising on the width of the `searching` and `showing` fields. In my opinion this also looks much better from a design perspective.

### Acceptance Criteria:
- Make the break point more generous 
- Ensure the magnifying icon within the field doesn't cover the placeholder text.

### Screenshots:
Screenshot 2022-07-06 at 15 43 25

### DIFF
--- a/ui/src/pages/search.tsx
+++ b/ui/src/pages/search.tsx
@@ -277,7 +277,7 @@ const Search: Types.NextPage<Props> = (props): React.ReactElement => {
                         <form
                             name="query-form"
                             id="query-form"
-                            className="col-span-12 lg:col-span-3 xl:col-span-4"
+                            className="col-span-12 lg:col-span-7"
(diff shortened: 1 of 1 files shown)


tomgiddingsjisc (approved):
LGTM

PR #187, opened 2022-07-07
OCT-247 - Ethical Statement blocking publication

CarlyRich:
The purpose of this PR was to fix a bug in the ethical statement section of the publication flow.
JIRA Ticket: https://jiscdev.atlassian.net/browse/OCT-247

Ethical statement doesn't count as completed if second option is selected
This is only for ‘Results’ types.

Incorrect functionality at the moment:

- When choosing the second option, the ‘ethical statement’ section is shown as not complete.
- The optional free text field is shown with both options, it should only be shown when the first option is selected.

I have also:
- Implemented accessibility fixes so that lighthouse ci passes
- Added in a missing doi to the test seed data so that api tests can be run locally

---

### Acceptance Criteria:

A user wants to publish a 'Results' publication and the results involved human or animal subjects, then: they tick 'The results in this publication involved human or animal subjects.' and a free text field appears so they can optionally add relevant info, the publication can be saved/published and the publication page shows this text and any information from the free text field. 

A user wants to publish a 'Results' publication and the results didn't involve human or animal subjects, then: they tick 'The results in this publication do not involve human or animal subjects.', the publication can be saved/published and the publication page shows this text.

A user does not pick an option from the 'Ethical statement' section of the publication flow. This section is shown with a warning on the 'review' page, and the publication cannot be published.

### DIFF
--- a/api/prisma/migrations/20220706145818_update_ethical_statement/migration.sql
+++ b/api/prisma/migrations/20220706145818_update_ethical_statement/migration.sql
@@ -1,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "Publication" ALTER COLUMN "ethicalStatement" SET DATA TYPE TEXT;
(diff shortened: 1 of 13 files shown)


tomgiddingsjisc (approved):
As discussed in standup, the product team are happy with the risk involved wrt unit test suite being broken.

LGTM!

PR #194, opened 2022-07-11
OCT-240 Data permissions statement blocking publication

CarlyRich:
The purpose of this PR was to fix a bug in the data permissions section of the publication flow.
JIRA Ticket: https://jiscdev.atlassian.net/browse/OCT-240

The mandatory free text field is appearing with both options but should only show when option 1 is chosen, so it is blocking the publication flow. 
I have also changed the logic so that the 'Data permissions statement provided by' section on the review page only shows when option 1 has been selected for data permissions, and corrected a few typos.

This is only for ‘Results’ types. 
Incorrect functionality at the moment:

- On the ‘review and publish’ the section ‘Data permissions statement provided by' is shown even when the second option for ‘Data permissions’ is selected so the free text field is not needed to continue with the publication.
- As this free text field is set as required even when the second option is selected, this is blocking the publication flow as the box is not shown to fill in when the second option is selected.
- Just to clarify, there should be no section shown on ‘review and publish’ page for Data Access statement as this is optional.

---

### Acceptance Criteria:

 A user wants to publish a 'Results' publication and the results does not involve access to materials owned or copyrighted materials (except those in the private ownership of the authors). They choose option two in the 'data permissions' section. On the review page, they cannot see the 'Data permissions statement provided by' section, and they can publish with no blockers from the data permissions section and the right data is sent to the database. The right information is shown on the 'preview' publication and the 'published' publication.

A user wants to publish a 'Results' publication and the results involved access to owned or copyrighted materials. They choose option one on the 'data permissions' section. The mandatory free text box appears and they do not enter anything. On the review page, they are stopped from publishing as the publish button is disabled.

A user wants to publish a 'Results' publication and the results involved access to owned or copyrighted materials. They choose option one on the 'data permissions' section. The mandatory free text box appears and they fill this in. On the review page, both sections are appearing as completed and they can publish with no blockers from the data permissions section and the right data is sent to the database. The right information is shown on the 'preview' publication and the 'published' publication.

### DIFF
--- a/ui/src/components/Publication/Creation/DataStatements.tsx
+++ b/ui/src/components/Publication/Creation/DataStatements.tsx
@@ -4,15 +4,13 @@ import * as OutlineIcons from '@heroicons/react/outline';
 
 import * as Components from '@components';
 import * as Stores from '@stores';
+import * as Config from '@config';
 
 const ethicalStatementOptions: string[] = [
     'The results in this publication involved human or animal subjects.',
     'The results in this publication do <strong>not</strong> involve human or animal subjects.'
 ];
-const dataPermissionsOptions: string[] = [
-    'The results in this publication involved access to owned or copyrighted materials.',
-    'The results in this publication do <strong>not</strong> involve access to materials owned or copyrighted materials (except those in the private ownership of the authors).'
-];
+
 const dataAccessOptions: string[] = [
     'This publication includes all the relevant results or data collected by the authors.',
     'This publication contains a sample of relevant results or data, with the full data publicly stored on another platform.',
@@ -107,7 +105,7 @@ const DataStatements: React.FC = (): React.ReactElement => {
             <div>
                 <Components.PublicationCreationStepTitle text="Data permissions statement" required />
                 <fieldset className="my-8 space-y-3">
-                    {dataPermissionsOptions.map((option) => (
+                    {Config.values.dataPermissionsOptions.map((option) => (
                         <label
                             key={option}
                             htmlFor={option}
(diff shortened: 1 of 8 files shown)


ui/src/components/Publication/Creation/DataStatements.tsx
```diff
@@ -4,15 +4,13 @@ import * as OutlineIcons from '@heroicons/react/outline';
...
 import * as Stores from '@stores';
+import * as Config from '@config';
 
 const ethicalStatementOptions: string[] = [
     'The results in this publication involved human or animal subjects.',
     'The results in this publication do <strong>not</strong> involve human or animal subjects.'
 ];
-const dataPermissionsOptions: string[] = [
-    'The results in this publication involved access to owned or copyrighted materials.',
-    'The results in this publication do <strong>not</strong> involve access to materials owned or copyrighted materials (except those in the private ownership of the authors).'
```
shaunajisc, line 14: It might be better to store this as a keyed object, as if we want to change the text in the future, previous data will be stored differently on the database but will mean the same thing :)

shaunajisc (changes requested):
Looks great - just a few comments! Also, it would be good to get some tests in there but I can see you have the different scenarios in the acceptance criteria for test to be built and I think the risk for no tests has been accepted?

CarlyRich:
@shaunajisc and I talked about having a discussion in the future on the structure e.g. at the moment we are using array of strings to map over for features like EthicalStatement, DataPermissions, DataAccess etc.

shaunajisc (approved):
LGTM ⭐️👍

PR #196, opened 2022-07-13
OCT-254 swap hard-coded config to use env vars

karmseveeratjisc:
Summary:

API needs to be updated to point to live URL, not int. 

 

for the API, for int it needs to be [int.octopus.ac](http://kint.octopus.ac/), on prod it needs to be [Octopus](http://www.octopus.ac/) 
For local environments it is not clear what make sense, localhost:3000 (but datacite might reject that)

There are additional places where a similar URL is constructed (based on the value of STAGE env var), we want these all to be managed by environment variables explicitly and not by STAGE. 

 

[email.t](http://emial.td/)s has 
baseURL constructed from STAGE
from address (again based on stage)

add new env vars:

EMAIL_SENDER_ADDRESS 
no-reply@www.octopus.ac

no-reply@int.octopus.ac

no-reply@local.ac

AUTHORISATION_CALLBACK_URL

https://www.octopus.ac/login

https://int.octopus.ac/login

https://localhost:3001/login

 

BASE_URL

https://www.octopus.ac

https://int.octopus.ac

https://localhost:3001

 

 

Note from Matt: This will need a ticket and a code change I think. It's hard coded to one environment, and to pull that out to be environment dependent is going to require a bit of a change through the code I think.

Acceptance Criteria:

API

[octopus/helpers.ts at OCT-128 · JiscSD/octopus](https://github.com/JiscSD/octopus/blob/OCT-128/api/src/lib/helpers.ts#L78) 

Risk assessment: Low


MattMcGowanJisc (approved):
Looks all good to me, good job.

No code change, but I think when we set the `EMAIL_SENDER_ADDRESS` on prod it'll need to be `no-reply@prod.octopus.ac`. This is just because SES is currently setup with the prod.octopus.ac domain set up. I think ultimately we should change this to `no-reply@octopus.ac`, but that's for a later change I think that will require reconfiguring SES and DKIM etc.

PR #197, opened 2022-07-14
Api env var consolidation

MattMcGowanJisc:
The purpose of this draft PR is to demo how we could remove the environment variables from .env files for releasing the API.

This appears to still work locally, but needs more testing by developers on their local machine.

This will remain a draft for discussion. If we want to take this approach, we'll create a ticket and put this through the proper process.


api/serverless.yml
```diff
@@ -23,9 +23,21 @@ provider:
...
         ELASTICSEARCH_ENDPOINT: ${ssm:/elasticsearch_endpoint_${self:provider.stage}_octopus}
+        JWT_SECRET: ${ssm:/jwt_secret_${self:provider.stage}_octopus}
+        VALIDATION_CODE_EXPIRY: 10
+        VALIDATION_CODE_ATTEMPTS: 3
+        ORCID_ID: ${ssm:/orcid_app_id_${self:provider.stage}_octopus}
+        ORCID_SECRET: ${ssm:/orcid_secret_key_${self:provider.stage}_octopus}
+        DOI_PREFIX: ${ssm:/doi_prefix_${self:provider.stage}_octopus}
+        DATACITE_ENDPOINT: ${ssm:/datacite_endpoint_${self:provider.stage}_octopus}
+        DATACITE_USER: ${ssm:/datacite_user_${self:provider.stage}_octopus}
+        DATACITE_PASSWORD: ${ssm:/datacite_password_${self:provider.stage}_octopus}
```
MattMcGowanJisc, line 39: All the environment variables for deployment are obtained from AWS parameter store, meaning a `.env` file is not required.

infra/modules/api/main.tf
```diff
@@ -0,0 +1,48 @@
...
+  name  = "datacite_user_${var.environment}_${var.project_name}"
+  // Must be manually created before terraform run. It is required for API releases.
+  // The username of the datacite user, used for minting DOIs.
+}
+
+data "aws_ssm_parameter" "datacite_password" {
+  name  = "datacite_password_${var.environment}_${var.project_name}"
+  // Must be manually created before terraform run. It is required for API releases.
+  // The password of the datacite user, used for minting DOIs.
+}
```
MattMcGowanJisc, line 48: The other environment variables added as `data` blocks. I think this is worth discussing on if we want to do this.

These data blocks make the terraform fail if the parameters don't exist in parameter store. This moves the failure from happening at API deploy time, back to infrastructure creation time. I think it also documents what's required.

However, this might not be where we want the failure to happen. Perhaps it just needs a multi-line comment in this file?

MattMcGowanJisc (commented):
Added comments to the files to explain rationale.

PR #200, opened 2022-07-15
Affiliation organisations updating across all drafts - bug - (OCT-256)

shaunajisc:
The purpose of this PR was to fix a bug where If you add or remove an affiliated organisation, it updates across all draft publications on: 
- Both Windows/Chrome and Mac/Safari
- Across multiple publication types. 
- When using both ror and manual input

The behaviour of the bug is as follows: 
- If I remove affiliation, it also disappears from other drafts
- If I update one of the drafts to something else, both drafts update
- If multiple are added, all appear across publications 
- Funder statement however seems to be working as intended.

Through testing and looking at the schema, I found that it isn't a problem with the backend, as affiliated organisations are being stored per `publicationId`. The bug is caused by the data persisting using global state, as when the author saves the publication the affiliations data isn't reset. I have resolved this by adding in `affiliations` and `affiliationsStatement` to the reset `function` within global store.


tomgiddingsjisc (approved):
LGTM - Well spotted on global state.

PR #204, opened 2022-07-18
Bug - Co-authors unable to approve Results publications (OCT-261) - UI

shaunajisc:
The purpose of this PR was to fix a bug where co-authors are unable to approve 'results' publications, as per this [Jira ticket.](https://jiscdev.atlassian.net/jira/software/projects/OCT/boards/836?selectedIssue=OCT-261)

During testing, in two cases co-authors were unable to approve results publications, as they came across the following error: 

image

Upon inspection, I found that the error is being thrown by a section within the publication page, where the `ethicalStatement` is being parsed, this is throwing an error as the data is a `string` not an `object`. This has been resolved by removing the `parse()` function from the line.


tomgiddingsjisc (approved):
This looks right, ethicalStatement is a selection of predefined plain text options now and as such does not require parsing.

LGTM!

PR #206, opened 2022-07-20
Conflict of interest - 'No' value consistently not being retained - BUG - (OCT-263)

shaunajisc:
The purpose of this PR was to fix undesired behaviour where the confilict of interest value isn't being retained when a user selects 'no', saves the draft and then returns to the publication, as per this [Jira ticket.](https://jiscdev.atlassian.net/jira/software/projects/OCT/boards/836?selectedIssue=OCT-263)

It was set up so that the `conflictOfInterestStatus` was either `true` or `false`, `false` by default on the backend and `true` on the UI. This was then overriding the value on the database if the user has selected `false`, so although the data was being retained the UI wasn't giving any feedback that this is the case. 

I have changed it so that the `conflictOfInterestStatus` now accepts three values: undefined, false and true. If a user hasn't selected any value, both radio fields will show as blank. I also needed to change checks within other stages as it wasn't differentiating between false and undefined.

### DIFF
--- a/api/prisma/migrations/20220720091606_conflict_of_interest_bug_fix/migration.sql
+++ b/api/prisma/migrations/20220720091606_conflict_of_interest_bug_fix/migration.sql
@@ -1,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "Publication" ALTER COLUMN "conflictOfInterestStatus" DROP DEFAULT;
--- a/api/prisma/schema.prisma
+++ b/api/prisma/schema.prisma
@@ -56,7 +56,7 @@ model Publication {
     type                               PublicationType
     title                              String?
     licence                            LicenceType            @default(CC_BY)
-    conflictOfInterestStatus           Boolean?               @default(false)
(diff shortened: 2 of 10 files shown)


CarlyRich:
Hey Shauna, when I have tested this it seems that COI is set originally as 'No', is that the correct functionality? As it means that a user can accidentally skip this stage and still publish.

shaunajisc:
Hey @CarlyRich, hmm that's strange - it's showing as both are unselected on my computer locally. Are you definitely on the right branch? 🙂

Screenshot 2022-07-22 at 10 22 23

CarlyRich (commented):
It worked after updating my node modules, all looks good! 
One comment would be that when selecting 'yes' and typing something, then changing to 'no', the text is still present and will be submitted. It won't show anywhere in the UI but it will be in the database so may be good to just add some code to clear the free text field when switching to no (I think ethical statement and data permissions do this too)

CarlyRich (approved):
LGTM 🏅

PR #210, opened 2022-07-26
Data access - 'other' field not being retained - BUG - (OCT-264)

shaunajisc:
The purpose of this PR was to fix a bug where if a user selects the 'other' the data isn't retained, along with any text the user inputs in the text field, as per this [Jira ticket.](https://jiscdev.atlassian.net/jira/software/projects/OCT/boards/836?selectedIssue=OCT-303)

This was being caused because the text field is being initialised locally as `''` and isn't referencing the global state, so although the data was being retained on the backend, the UI within the build process was giving no feedback that this was the case. I have now amended it so that once the page mounts, we check to see if the dataAccessStatement matches any of the other statements, if not the `other` field is selected and the free text field is populated with the `dataAccessStatement`.


CarlyRich (approved):
Some work can be done separately to this on how we can make this section more user-friendly and align to how the other sections are done, e.g. show free text box dynamically and have this field as mandatory when user selects 'Other'. 
All good for this PR, fixes this bug 👍 !

karmseveeratjisc (approved):
I'll go with Carly's greater knowledge and approve, looks good

PR #211, opened 2022-08-01
Make buttons more prominent (OCT-243) - UI

shaunajisc:
The purpose of this PR was to improve the UX of the publication process:
- Add a 'save' button to the bottom of every page
- Add a 'preview' button to the bottom of the 'review' page
- Add teal styling to 'preview' and 'publish' buttons
- Add green styling to 'save' button
- Add message to each step of publication flow - "Remember to save this draft before navigating away from the publication form."
- Changed where coauthor and affiliation button are located to a more noticeable position

### DIFF
--- a/ui/src/layouts/BuildPublication.tsx
+++ b/ui/src/layouts/BuildPublication.tsx
@@ -17,18 +17,21 @@ type NavigationButtonProps = {
     className?: string;
     icon: React.ReactElement;
     iconPosition: 'LEFT' | 'RIGHT';
+    saveButton?: boolean;
 };
 
 const NavigationButton: React.FC<NavigationButtonProps> = (props) => (
     <button
         disabled={props.disabled}
         onClick={props.onClick}
-        className={`font-lg flex items-center space-x-2 rounded-sm py-1 text-grey-800 outline-none transition-colors duration-500 focus:ring-2 focus:ring-yellow-400 disabled:opacity-50 disabled:hover:cursor-not-allowed dark:text-white-50 ${
+        className={`font-lg flex items-center  rounded-xl ${
+            props.saveButton ? 'bg-green-600' : 'bg-teal-700'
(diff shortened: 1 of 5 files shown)


ui/src/layouts/BuildPublication.tsx
```diff
@@ -353,14 +357,31 @@ const BuildPublication: React.FC<BuildPublicationProps> = (props) => {
                     <div className="mb-12">{props.children}</div>
                     {/* bottom next and back nav buttons */}
                     <div className="flex justify-end space-x-8 ">
+                        <NavigationButton
```
CarlyRich, line 360: Is this preview button meant to be just on the bottom of the 'review' page, or on every step of the publication flow?

CarlyRich (changes requested):
A few comments, I don't think we want to change all buttons across the site? As other buttons may have different spacing where having the background a different colour may not work. And just on whether the preview button is needed on all of the publication step pages (as there would be two preview buttons on each page then), I think from AC it was meant for just the review page?

ui/src/components/Publication/Creation/DataStatements.tsx
```diff
@@ -43,7 +43,15 @@ const DataStatements: React.FC = (): React.ReactElement => {
     // Data access statement
```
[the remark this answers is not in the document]
shaunajisc: Ah - my bad!

ui/src/layouts/BuildPublication.tsx
shaunajisc: Every step - and grayed out when the publication hasn't been fully filled out. Similar to the bar above the publication flow

CarlyRich: Ah okay then :) if a second preview button is being added to every publication step page then it might be good to update the acceptance criteria in this PR to match that as currently is says 'Add Preview button to bottom of review page'

AshleyRedman:
cc: https://github.com/JiscSD/octopus/discussions/213

CarlyRich (approved):
lgtm 👍

PR #212, opened 2022-08-01
Broken co-author link - (OCT-295) - BUG

shaunajisc:
The purpose of this PR was to fix a broken co author link on the 'view a publication' page. The link was using the user's ORCHID id in the path params for the URL, instead of the `linkedUser` id. I have now amended it so it uses the `linkedUser` id.

### DIFF
--- a/ui/src/lib/types.ts
+++ b/ui/src/lib/types.ts
@@ -71,7 +71,7 @@ export type PublicationCreationStoreType = {
     ethicalStatementFreeText: string | null;
     updateEthicalStatementFreeText: (ethicalStatementFreeText: string | null) => void;
     updateEthicalStatement: (ethicalStatement: string) => void;
-    dataAccessStatement: string | null;
+    dataAccessStatement: string;
     updateDataAccessStatement: (dataAccessStatement: string | null) => void;
     dataPermissionsStatement: string | null;
     updateDataPermissionsStatemnt: (dataPermissionsStatement: string) => void;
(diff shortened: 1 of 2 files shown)


ui/src/components/Publication/Creation/DataStatements.tsx
```diff
@@ -43,7 +43,15 @@ const DataStatements: React.FC = (): React.ReactElement => {
     // Data access statement
     const dataAccessStatement = Stores.usePublicationCreationStore((state) => state.dataAccessStatement);
     const updateDataAccessStatement = Stores.usePublicationCreationStore((state) => state.updateDataAccessStatement);
-    const [dataAccessStatementOther, setDataAccessStatementOther] = React.useState('');
+    const [dataAccessStatementOther, setDataAccessStatementOther] = React.useState<string | null>('');
```
karmseveeratjisc, line 46: this seems to revert a change elsewhere and also is unrelated to the fix

ui/src/lib/types.ts
```diff
@@ -71,7 +71,7 @@ export type PublicationCreationStoreType = {
     ethicalStatementFreeText: string | null;
     updateEthicalStatementFreeText: (ethicalStatementFreeText: string | null) => void;
     updateEthicalStatement: (ethicalStatement: string) => void;
-    dataAccessStatement: string | null;
+    dataAccessStatement: string;
```
karmseveeratjisc, line 74: again this is not related to the fix (and is already merged, I find github a bit strange in that respect bitbucket would show this as either not a change or a conflict)

CarlyRich:
@karmseveeratjisc If you want to re-review please :) thanks!

karmseveeratjisc (approved):

PR #220, opened 2022-08-15
Write a linked publication / review - OCT-200 (UI)

shaunajisc:
The purpose of this PR was to create a link for logged in users to write a linked publication from any publication page. 

---

### Acceptance Criteria:

- There is a ‘write a linked publication’ within the actions section on each publication page. 
- As with ‘write a review’. only logged in users can view this option. 
- After discussions with Lola and Alex, we have decided to give users alternative options to write either linked publication types within the publication sidebar (eg. "Write a linked Hypothesis" or "Write a linked Problem")
- The draft publication will then also have the ‘Linked publications’ section pre-populated to match the publication from which the draft was created (i.e. the publication where the user clicked ‘write a linked publication’
- Link can be deleted. 
- Additional links can be added (within usual publication chain rules).

The linked publication types are as follows: 
Screenshot 2022-08-15 at 14 42 01

---

### Screenshots:
Screenshot 2022-08-15 at 14 43 21


ui/src/lib/helpers.tsx
```diff
@@ -282,3 +282,17 @@ export const checkLinkIsValid = (text: string) => {
...
+        PROBLEM: ['PROBLEM', 'HYPOTHESIS'],
+        HYPOTHESIS: ['PROBLEM', 'PROTOCOL'],
+        PROTOCOL: ['PROBLEM', 'DATA'],
+        DATA: ['PROBLEM', 'ANALYSIS'],
+        ANALYSIS: ['PROBLEM', 'INTERPRETATION'],
+        INTERPRETATION: ['PROBLEM', 'REAL_WORLD_APPLICATION'],
+        REAL_WORLD_APPLICATION: ['PROBLEM']
+    };
+
+    return types[publicationType as keyof typeof types];
```
florin-holhos, line 297: Your first attempt with the linkedPublicationTypes as an object was better. You just needed to provide the type for it.

For example:

export const linkedPublicationTypes: { [key: string]: Types.PublicationType[] } = {
        PROBLEM: ['PROBLEM', 'HYPOTHESIS'],
        HYPOTHESIS: ['PROBLEM', 'PROTOCOL'],
        // ...rest of them
};

And then in SidebarCard/Actions: Helpers.linkedPublicationTypes[props.publication.type].map(...)  will work as expected.

florin-holhos (dismissed):

ui/src/lib/helpers.tsx
shaunajisc: Ah ok - I changed it because I noticed that's how the `helpers.formatPublicationType` does it and wanted to keep it in line with that, just incase we need to reuse the linked publication types helper somewhere else in the code then there's no confusion. Shall I change it back to how it was?

florin-holhos: Oh, I see. Then you can change the type of types and simply return types[publicationType];

shaunajisc: Ah ok - good point! Thanks - I've changed it back to the object :)

florin-holhos (approved):

PR #225, opened 2022-08-22
OCT-324

Harry-Jisc:
The purpose of this PR was to fix the dark mode issues that affected the main text editor on publications. 

### Tests:

I've mostly focused on fixing what was raised in the ticket and anything else I could find that broke, try and use the features in the editor to look for anything else that could have styling conflicts

To recreate the issue follow these steps:

1. Open draft publication
2. Add main text with table and link formatting (templates available)
3. Apply dark mode
4. Check styling of tables and links (see screenshots)

### Screenshots:

**Before**
old-dark-mode

**New**
new-dark-mode


shaunajisc (dismissed):
LGTM 🙂👍

ui/src/config/values.ts
```diff
@@ -897,3 +897,32 @@ export const HTMLStyles = `
...
+    focus:prose-a:bg-yellow-300 
+    prose-blockquote:border-teal-500 
+    prose-blockquote:text-grey-600
+    prose-strong:text-grey-800
+`
+    .split(' ')
+    .map((i) => {
+        return i.replace(/\n/g, '').length ? i.replace(/\n/g, '') : undefined;
+    })
+    .filter((i) => !!i)
```
florin-holhos, line 927: Is the map and filter really needed for HTMLStyles and HTMLStylesTiptapEditor?

Harry-Jisc: I can remove the filter without error and still achieves the same result, otherwise when the map is gone it errors with: 
```
InvalidCharacterError: DOMTokenList.add: The token can not contain whitespace.
```

florin-holhos: Did you try .split('\n').join(' ')?

Harry-Jisc: That works, is it work removing the filter/map from HTMLStyles as well, since from what I understand, that's what it's attempting to achieve anyway, albiet out of the scope of this ticket?

florin-holhos: Yes, there are lots of bits of code like this that can be improved on the fly when we're working in a specific area and we see them.

florin-holhos (approved):

ui/src/config/values.ts
karmseveeratjisc: split/join does however leave unnecessary whitespace in the resulting string so isn't as neat, a reaplce is slighlt better but leaves a leading space:

.replaceAll(/\s+/g,' ')

karmseveeratjisc: the whitespace actually makes no difference in this case

PR #229, opened 2022-08-25
OC-745: Generate separate sitemaps per 50k publications on back end

florin-holhos:
The purpose of this PR was to generate a dynamic sitemap, including publications.

---

### Acceptance Criteria:

- Every 24 hours, a cron job triggers a lambda to generate a sitemap for publications which is then stored in an s3 bucket 
- An API endpoint is present which returns links to XML sitemaps for publications
- The sitemaps at /publications/[number].xml in an s3 bucket contain links to the “[number]th” 50,000 publications in octopus (a consistent sort order between the sitemaps is needed but it doesn’t especially matter what it is). For example, /publications/3.xml would contain the 100,001st publication up to the 150,000th publication.


finlay-jisc:
Closing this pull request to make a new one with the same branch so that Florin can review it.

PR #232, opened 2022-08-30
OCT-350 Rich text editor - field and table

florin-holhos:
The purpose of this PR was to replace References textarea with a rich text editor and also update the references table to display html content

---

### Acceptance Criteria:

As per [OCT-350](https://jiscdev.atlassian.net/browse/OCT-350)

### Tests:

Will be added in a new commit

### Screenshots:

image

image

### DIFF
--- a/ui/src/components/Publication/Creation/MainText.tsx
+++ b/ui/src/components/Publication/Creation/MainText.tsx
@@ -1,15 +1,131 @@
-import React from 'react';
-
+import React, { useState } from 'react';
 import * as api from '@api';
-
 import * as Assets from '@assets';
 import * as Components from '@components';
-import * as Config from '@config';
 import * as Helpers from '@helpers';
 import * as OutlineIcons from '@heroicons/react/outline';
 import * as Interfaces from '@interfaces';
 import * as Stores from '@stores';
 import * as Types from '@types';
+import * as tiptap from '@tiptap/react';
+import * as FAIcons from 'react-icons/fa';
+import * as Config from '@config';
+
+import StarterKit from '@tiptap/starter-kit';
+import Underline from '@tiptap/extension-underline';
+
+const menuIconStyles = 'p-2 hover:bg-grey-100 hover:rounded focus:outline-yellow-500';
+const activeMenuIconStyles = 'p-2 bg-grey-100 rounded focus:outline-yellow-500';
+
+interface MenuBarProps {
+    editor: tiptap.Editor;
+    loading: boolean;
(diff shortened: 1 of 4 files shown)


Harry-Jisc:
Don't think this is something we need to add to this branch specifically, but I just added 44 references in one go, **which it picked up and displayed perfectly**, however it took about 10-12 seconds ish to do it and forces the screen to the bottom of the page, it might be worth thinking about a a modal that states 'Adding references' etc at some point?

Harry-Jisc:
When adding with two lines between a reference, the table does not appear until refresh and after refresh anything after the double line does not appear.

[double line](https://gyazo.com/6a4b37cbdf1c1c87ce18eb3b63d227e0)

As you can see the final reference was cut off from the table
image

Is something to be fixed or something the users need to be trained on since, it does state " References should be line-separated" which implies single line?

karmseveeratjisc:
the order is changing too...

think we said either we need to detect blank references or we need to insert them as blank entries and the user delete

ui/src/components/Publication/Creation/MainText.tsx
```diff
@@ -47,68 +161,81 @@ const MainText: React.FC = (): React.ReactElement | null => {
...
+
+                for (let i = 0; i < paragraphsArray.length; i++) {
+                    const currentParagraph = paragraphsArray[i].trim();
+                    const paragraphElement = new DOMParser()
+                        .parseFromString(currentParagraph, 'text/html')
+                        .querySelector('p');
+                    const textContent = paragraphElement?.textContent?.trim();
+
+                    if (!textContent) {
+                        return;
```
Harry-Jisc, line 181: Instead of returning here, if this becomes `continue;` instead, it keeps the order the allows for the double lines without inserting spaces, or at least it did for me 😄

florin-holhos: yeah, I can't remember exactly why I added this return right here. The UI is updating accordingly now. Good catch! 😄

florin-holhos:
> Don't think this is something we need to add to this branch specifically, but I just added 44 references in one go, **which it picked up and displayed perfectly**, however it took about 10-12 seconds ish to do it and forces the screen to the bottom of the page, it might be worth thinking about a a modal that states 'Adding references' etc at some point?

This will be fixed in [OCT-360](https://jiscdev.atlassian.net/browse/OCT-360)

Harry-Jisc (approved):
With the changes to ordering/spacing sorted and comments resolved I'm happy to approve! 👍

PR #236, opened 2022-09-02
Generate a unique ID on the frontend and change the logic so references are stored in state on the frontend until a user saves their publication

shaunajisc:
The purpose of this PR was to amend the references feature so that we generate the unique ID on the frontend, which then enables us to store the references in global state on the frontend and to then send them to the database when a user saves their draft, as per this [Jira ticket](https://jiscdev.atlassian.net/jira/software/projects/OCT/boards/836?selectedIssue=OCT-360).

This PR contains changes to the frontend, global state, a new `updateAll` references endpoint with the following path: 
```
PATCH publications/${publicationId}/reference
```
that takes the following body: 
```JSON
[{
    "id": "uniqueCUID",
    "publicationId": "publication-user-1-problem-2-draft",
    "type": "TEXT",
    "text": "<p>https://github.com/JiscSD/octopus/compare/unique-id?expand=1</p>",
    "location": null
}]
```

and a `removeAll` endpoint that has the following path: 
```
DELETE publications/${publicationId}/reference
```

in addition to changes to the schema so that the database no longer generates the id and accepts the new id in the body of the request and we have also added middleware to verify that a user is the author to all of the relevant reference endpoints.

---

### Tests:

We will need to write tests for all of the endpoints in references, we don't have any as it stands


ui/src/lib/api.ts
```diff
@@ -60,7 +60,7 @@ export const patch = async (url: string, body: Interfaces.JSON, token?: string |
     return response;
 };
 
-export const put = async (url: string, body: Interfaces.JSON, token: string | undefined): Promise<AxiosResponse> => {
+export const put = async (url: string, body: any, token: string | undefined): Promise<AxiosResponse> => {
```
karmseveeratjisc, line 63: can we not keep this more specific?

karmseveeratjisc: to be honest the api.js uses a pattern I'm not massively keen on (why does UI specific code have URLs for the API in it, why not have specific methods here? e.g putReferences, putPublication, seems lazy) so perhaps we dont address here

shaunajisc: We were having trouble getting the `Interfaces.JSON` to accept an array instead of an object so we used `any`, I've just checked to see if we can do it a different way but it's still rejecting it. But no problem we can always revisit in the future 🙂

florin-holhos: this will be changed. we just forgot it when we tested the update. There is more stuff into the PR that will be removed

api/src/components/publication/controller.ts
```diff
@@ -136,6 +137,15 @@ export const create = async (
 
         const publication = await publicationService.create(event.body, event.user, doi);
 
+        // add references to the publication
+        event.body.references?.map(async (item) => {
```
florin-holhos, line 141: Is this change still needed? I think we can remove this since we have dedicated endpoint for references now.

api/src/components/publication/controller.ts
```diff
@@ -201,6 +211,16 @@ export const update = async (
 
         const updatedPublication = await publicationService.update(event.pathParameters.id, event.body);
 
+        // updating references
+        // if references already exist, then we need to rewrite them instead of appending them to the array
+        if (publication?.references !== event.body.references) {
```
florin-holhos, line 216: Same as above here. We don't need this anymore

api/src/lib/interface.ts
```diff
@@ -83,6 +83,7 @@ export interface CreatePublicationRequestBody {
     dataAccessStatement?: string;
     selfDeclaration?: boolean;
     affiliationStatement?: string;
+    references?: Reference[];
```
florin-holhos, line 86: I think we can remove this since we have separate endpoints for the references. We update the references for a particular publication separately

api/src/lib/interface.ts
```diff
@@ -124,6 +125,7 @@ export interface UpdatePublicationRequestBody {
     dataPermissionsStatementProvidedBy?: string;
     dataAccessStatement?: string;
     selfDeclaration?: boolean;
+    references?: Reference[];
```
florin-holhos, line 128: is this still needed?

ui/src/components/Publication/Creation/MainText.tsx
```diff
@@ -141,120 +142,72 @@ const MainText: React.FC = (): React.ReactElement | null => {
...
-        if (publicationId) {
-            try {
-                const response = await api.get(`/publications/${publicationId}/reference`, user?.token);
-                updateReferences(response.data);
-            } catch (err) {
-                /**
-                 * @TODO - handle errors - eg. cannot load references...
-                 */
-            }
+    const [loading, setLoading] = React.useState(false);
```
florin-holhos, line 145: This loading state is obsolete from what I saw. Same as const user = ... above. They are not used anymore

ui/src/lib/types.ts
```diff
@@ -95,7 +95,15 @@ export type PublicationCreationStoreType = {
...
-export type JSONValue = string | number | boolean | { [x: string]: JSONValue } | Array<JSONValue> | null | undefined;
+export type JSONValue =
+    | string
+    | number
+    | boolean
+    | { [x: string]: JSONValue }
+    | Array<JSONValue>
+    | null
+    | undefined
+    | Interfaces.Reference[];
```
florin-holhos, line 106: Interfaces.Reference[] can be removed as it doesn't help in this case. We will find a better way for this

ui/src/pages/publications/[id]/edit.tsx
```diff
@@ -3,6 +3,7 @@ import * as Framer from 'framer-motion';
 import Head from 'next/head';
 import * as Router from 'next/router';
 import React from 'react';
+import useSWR from 'swr';
```
Comment (florin-holhos) [inline ui/src/pages/publications/[id]/edit.tsx:6]: this import can be removed if not used

api/src/middleware/checkOwnership.ts
```diff
@@ -0,0 +1,49 @@
...
+            // if there's no user account, and authentication is *not* optional, then the request is blocked.
+            if (!user) {
+                return response.json(401, { message: 'Please enter either a valid apiKey or bearer token.' });
+            }
+            const publicationId = request.event.pathParameters?.id;
+
+            if (publicationId) {
+                const publication = await publicationService.get(publicationId);
+                if (publication && publication.createdBy !== user.id) {
+                    return response.json(401, {
```
karmseveeratjisc, line 32: I think this should be a 403 not a 401 (as authorization has been supplied but access to the resource has been denied)

api/src/components/reference/routes.ts
```diff
@@ -11,15 +11,31 @@ export const create = middy(referenceController.create)
...
     .use(middleware.httpJsonBodyParser())
     .use(middleware.authentication())
+    .use(middleware.checkOwnership())
     .use(middleware.validator(referenceSchema.update, 'body'));
 
+export const updateAll = middy(referenceController.updateAll)
+    .use(middleware.doNotWaitForEmptyEventLoop({ runOnError: true, runOnBefore: true, runOnAfter: true }))
+    .use(middleware.httpJsonBodyParser())
+    .use(middleware.authentication())
+    .use(middleware.checkOwnership())
```
karmseveeratjisc, line 28: I wonder if there's a better way of doing this,  - checkownership also checks the authentication so having both is duplication

looking at checkowner  you can use the user object in the request.event that authentication adds perhaps?

api/src/middleware/checkOwnership.ts
```diff
@@ -0,0 +1,49 @@
...
+            if (publicationId) {
+                const publication = await publicationService.get(publicationId);
+                if (publication && publication.createdBy !== user.id) {
+                    return response.json(403, {
+                        message: 'User is not the author of this publication.'
+                    });
+                }
+            }
+
+            Object.assign(request.event, { user });
```
karmseveeratjisc, line 38: don't think this is needed

api/serverless.yml
```diff
@@ -345,52 +345,13 @@ functions:
...
-        handler: src/components/reference/routes.create
+    updateAllReference:
+        handler: src/components/reference/routes.updateAll
         warmup: true
         events:
             - http:
                   path: ${self:custom.versions.v1}/publications/{id}/reference
-                  method: post
-                  cors: true
-    updateReference:
```
karmseveeratjisc, line 387: I'm now confused, why has this gone? we still have update in the code, and where is removeAll?

florin-holhos (approved):

ui/src/components/Publication/Creation/MainText.tsx
```diff
@@ -358,22 +300,18 @@ const MainText: React.FC = (): React.ReactElement | null => {
             </div>
 
             <div>
-                <Components.PublicationCreationStepTitle text="Short description" />
+                <Components.PublicationCreationStepTitle text="Description" />
```
karmseveeratjisc, line 303: this feels unrelated

karmseveeratjisc: yep it reverts 

https://github.com/JiscSD/octopus/commit/b169b11b12d5b30807d26dad6de09f97710e613d

ui/src/pages/publications/[id]/edit.tsx
```diff
@@ -193,6 +194,23 @@ const Edit: Types.NextPage<Props> = (props): React.ReactElement => {
     const [currentStep, setCurrentStep] = React.useState(defaultStep);
     const [publication] = React.useState(props.draftedPublication);
 
+    const fetchAndSetReferences = React.useCallback(async () => {
+        if (props.draftedPublication.id) {
+            try {
+                const response = await api.get(`/publications/${props.draftedPublication.id}/reference`, props.token);
+                updateReferences(response.data);
+            } catch (err) {
+                /**
```
Comment (karmseveeratjisc) [inline ui/src/pages/publications/[id]/edit.tsx:203]: we need to at least log something here

karmseveeratjisc (approved):

PR #241, opened 2022-09-06
Delete all references button - UI

shaunajisc:
The purpose of this PR was to create a button which deletes all references. 

I have built it so that it overwrites whatever's in `updateReferences` state, and only clears what's in the database upon the user saving the publication, similar to the functionality with the `addReferences` feature. 

---

### Screenshots:

https://user-images.githubusercontent.com/100135505/188662969-2071d90b-02b9-4027-80ed-aafa3aebf805.mov


florin-holhos (approved):

Harry-Jisc (commented):
The AC on this ticket states:

> This opens a warning message in modal view: 
> Are you sure you want to delete all references from this publication? This action cannot be undone. 
> Delete references/Cancel. 

but when clicking delete all references, it just removes them?

PR #250, opened 2022-09-14
OCT-378 User cannot login if ORCID profile returns nulls for optional fields

Harry-Jisc:
The purpose of this PR was to fix an issue where we ran the .find function on a null field, this now only runs if something is there, otherwise returns null when there are no external ids. 

---

### Acceptance Criteria:

- Bug is fixed
- User with the specified ORCID is able to login.
---

### Tests:
Detailed steps to reproduce:

Only user can reproduce, using ORCID 0000-0002-4299-8978, it should be possible to temporarily swap the code that fetches the user information to use that ORCID instead of the logged in user’s.

---

### DIFF
--- a/api/src/components/authorization/controller.ts
+++ b/api/src/components/authorization/controller.ts
@@ -65,8 +65,8 @@ export const authorize = async (event: I.APIRequest<I.AuthorizeRequestBody>): Pr
         // eslint-disable-next-line @typescript-eslint/dot-notation
         const works = userInformation['activities-summary'].works['group'].map((work) => ({
             title: work['work-summary'][0].title.title.value || null,
-            doi: work['work-summary'][0]['external-ids'] === null ? null :
-                work['work-summary'][0]['external-ids']['external-id'].find(
+            doi:
+                work['work-summary'][0]['external-ids']?.['external-id'].find(
                     (externalId) => externalId['external-id-type'] === 'doi'
                 )?.['external-id-value'] || null,
             publishedDate: {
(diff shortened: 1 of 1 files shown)


api/src/components/authorization/controller.ts
```diff
@@ -65,7 +65,7 @@ export const authorize = async (event: I.APIRequest<I.AuthorizeRequestBody>): Pr
         // eslint-disable-next-line @typescript-eslint/dot-notation
         const works = userInformation['activities-summary'].works['group'].map((work) => ({
             title: work['work-summary'][0].title.title.value || null,
-            doi:
+            doi: work['work-summary'][0]['external-ids'] === null ? null :
```
florin-holhos, line 68: There's no need to check for null, you can use optional chaining for all the nested properties and put a "|| null" at the end.

From what I remember, we agreed to use optional chaining for all these mappings until we refine the login flow, because there's more stuff that can break. Any thoughts @karmseveeratjisc ??

karmseveeratjisc: ideally yes

Harry-Jisc: I have updated to use optional chaining for that particular mapping, what is meant by all these mappings ?

karmseveeratjisc: all the other places in that same file we do similar (accessing nested properties) , even for that one we could go mad and put something like

work['work-summary']?.[0]?.['external-ids']?.['external-id]

I think, as according to the spec external-ids is also optional

e.g. line 67 could be

work['work-summary']?.[0]?.title?.title?.value || null,

florin-holhos (approved):

PR #252, opened 2022-09-15
API tests for all of the references endpoints (OCT-376)

shaunajisc:
The purpose of this PR was to create tests for all of the endpoints within the references feature.

A few tests were passing when they should have been failing (eg. you could leave required fields in the `createReference` body blank or change the id/ publication id and you could delete references from a live publication) so this PR also contains changes where I added in logic to the controller to prevent this. 

---

### Tests:

```
 PASS  src/components/reference/__tests__/createReference.test.ts (16.503 s)
  create a reference
    ✓ User can add a reference to their DRAFT publication (1075 ms)
    ✓ User cannot add a reference to their LIVE publication (910 ms)
    ✓ User cannot add a reference to another users DRAFT publication (827 ms)
    ✓ User cannot add a reference to another LIVE publication (803 ms)
    ✓ User must send correct information to create a reference (no id) (885 ms)
    ✓ User must send correct information to create a reference (no publication id) (890 ms)
    ✓ User must send correct information to create a reference (incorrect type) (866 ms)
    ✓ User must send correct information to create a reference (no text) (891 ms)
```

```
 PASS  src/components/reference/__tests__/deleteAllReferences.ts (12.789 s)
  delete all references
    ✓ User can delete all references from their own draft publication (1206 ms)
    ✓ User must be the author of the publication (835 ms)
    ✓ The author can only delete references from a live publication (854 ms)
```

```
 PASS  src/components/reference/__tests__/deleteReference.test.ts (12.719 s)
  delete a reference
    ✓ User can delete a reference from their own draft publication (1221 ms)
    ✓ User must be the author of the publication (807 ms)
    ✓ The author can only delete references from a live publication (833 ms)
```

```
 PASS  src/components/reference/__tests__/getReference.test.ts (10.731 s)
  get references
    ✓ Any user (unauthenticated) can get the references for a live publication (1079 ms)
```

```
 PASS  src/components/reference/__tests__/updateAllReferences.test.ts (13.303 s)
  update all references
    ✓ User can update all references from their own draft publication (1078 ms)
    ✓ User must be the author of the publication to update the references (806 ms)
    ✓ The author can only update the references for a live publication (897 ms)
    ✓ All of the references should be replaced with the new references  (898 ms)
```

```
 PASS  src/components/reference/__tests__/updateReference.test.ts (14.067 s)
  update a reference
    ✓ An authenticated author can update a reference (1135 ms)
    ✓ The publication must be in draft state (896 ms)
    ✓ The user must be the author of the publication (840 ms)
    ✓ The id cannot be amended (815 ms)
    ✓ The publication id cannot be amended (802 ms)
```


api/src/components/reference/__tests__/deleteAllReferences.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    test('User must be the author of the publication', async () => {
+        const reference = await testUtils.agent
+            .delete('/publications/publication-interpretation-draft/reference')
+            .query({ apiKey: '987654321' });
+
+        expect(reference.status).toEqual(403);
+    });
+
+    test('The author can only delete references from a live publication', async () => {
```
florin-holhos, line 25: I think you meant "draft" publication

api/src/lib/interface.ts
```diff
@@ -434,7 +434,7 @@ export interface UpdateReferencePath {
 }
 
 export interface RemoveAllReferencesPath {
-    publicationId: string;
+    id: string;
```
florin-holhos, line 437: I think "publicationId" was a better naming here since we use this interface in the references controller and the publication is not the main resource of this endpoint.
For example: 
- "id" param would mean a Reference ID in the references controller
and
- "id" param would mean a Publication ID in the publications controller

shaunajisc: Good point, although I decided to change this to `id` as in the severless.yml file the route is `/publications/{id}/reference` (similar to the other routes eg. createReference/ updateReference etc). The interfaces file for the `removeAll` endpoint was `publicationId`, so the endpoint was incorrectly letting us add references to a live publication, as well as letting us add references to publications that don't exist. 

So to change this name to `publicationId` I'd need to change the route in the serverless.yml file to `/publications/{publicationId}/reference` which would be inconsistent with all of the other endpoints, and arguably not RESTful as it's implicit in the route that this is the id of the publication because it comes after the /publication/.

Hopefully that makes sense!

Agree with the interfaces naming, I'll change that now!

api/src/components/reference/__tests__/createReference.test.ts
```diff
@@ -0,0 +1,122 @@
...
+                text: '<p>Reference 1</p>',
+                type: 'TEXT'
+            });
+
+        expect(reference.status).toEqual(403);
+    });
+    test('User cannot add a reference to another users DRAFT publication', async () => {
+        const reference = await testUtils.agent
+            .post('/publications/publication-problem-draft/reference')
+            .query({ apiKey: '987654321' })
```
karmseveeratjisc, line 41: a comment on the existing tests that these test don't address (we might have to add to backlog I think) - the apiKey is a work around really (and only used by 1 user - octopus), it does not really represent a true 'user' (as that uses a decoded Bearer token), we can argue their equivalence but all the API tests for users will give me less confidence in what they are testing and also what they might be missing...

api/src/components/reference/__tests__/updateAllReferences.test.ts
```diff
@@ -0,0 +1,76 @@
...
+                {
+                    id: '04',
+                    location: 'http://octopus.ac',
+                    publicationId: 'publication-problem-draft',
+                    text: '<p>Reference 1</p>',
+                    type: 'TEXT'
+                }
+            ]);
+
+        expect(reference.body.count).toEqual(1);
```
karmseveeratjisc, line 74: I'd expect some checks of the returned data here to show it matches what was sent

karmseveeratjisc: actually, how is this different to the first test?

karmseveeratjisc: ok, to expand, each test starts with the same set of data so if the publication already has some references the first test and this test do the same (except one checks the status and one the count of results)

what you might want to do here is two consecutive update calls and check the data in the response changes

you might also want to update more than one reference in some places..

api/src/components/reference/controller.ts
```diff
@@ -14,7 +15,29 @@ export const get = async (event: I.AuthenticatedAPIRequest<undefined, undefined,
...
+        }
+
+        //check that the publication is live
+        if (publication.currentStatus !== 'DRAFT') {
+            return response.json(403, {
+                message: 'You can only add references to a draft publication.'
+            });
+        }
+
+        if (event.body.publicationId !== event.pathParameters.id || event.body.id == '' || event.body.text == '') {
```
karmseveeratjisc, line 34: ok this is part of what i created OCT-347 for, but both update and updateall need to do the same checks

api/src/components/reference/__tests__/updateReference.test.ts
```diff
@@ -0,0 +1,83 @@
...
+    });
+
+    test('The publication id cannot be amended', async () => {
+        const reference = await testUtils.agent
+            .patch('/publications/publication-interpretation-draft/reference/01')
+            .query({ apiKey: '000000005' })
+            .send({
+                id: '01',
+                location: 'http://octopus.ac',
+                publicationId: 'made-up-publication-id',
```
karmseveeratjisc, line 76: I think this only works because 'made-up-publication-id' does not exist (so prisma/db fails), if you used  'publication-problem-draft' it would update it,  incorrectly

karmseveeratjisc: actually it's the ownership code that is being hit, if you log the reference returned you see 

text: '{"message":"User is not the author of this publication."}',

api/src/components/reference/__tests__/deleteReference.test.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    test('User must be the author of the publication', async () => {
+        const reference = await testUtils.agent
+            .delete('/publications/publication-interpretation-draft/reference/01')
+            .query({ apiKey: '987654321' });
+
+        expect(reference.status).toEqual(403);
+    });
+
+    test('The author can only delete references from a live publication', async () => {
```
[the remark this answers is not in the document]
shaunajisc: My bad - I meant a draft publication

api/src/components/reference/__tests__/updateAllReferences.test.ts
shaunajisc: ok I've added this in the test 👍

api/src/components/reference/__tests__/deleteReference.test.ts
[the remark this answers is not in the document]
florin-holhos: As Mark said, 'The author can only delete references from a DRAFT publication', not LIVE publication.

api/src/components/reference/__tests__/updateAllReferences.test.ts
```diff
@@ -0,0 +1,88 @@
...
+                    publicationId: 'publication-problem-draft',
+                    text: '<p>Reference 1</p>',
+                    type: 'TEXT'
+                }
+            ]);
+
+        expect(reference.status).toEqual(403);
+    });
+
+    test('The author can only update the references for a live publication', async () => {
```
florin-holhos, line 43: DRAFT publication

api/src/components/reference/__tests__/updateReference.test.ts
```diff
@@ -0,0 +1,83 @@
...
+            .query({ apiKey: '000000005' })
+            .send({
+                id: '0000004',
+                location: 'http://octopus.ac',
+                publicationId: 'publication-problem-draft',
+                text: '<p>Reference 1</p>',
+                type: 'TEXT'
+            });
+
+        expect(reference.status).toEqual(403);
```
florin-holhos, line 66: I think this is returning 403 because of the API key (this user is not the author), not because of id field

api/src/components/reference/__tests__/updateReference.test.ts
florin-holhos: This is returning 403 because of the API key(this user is not the author). Not really testing if the publicationId can be changed or not.

To be honest, I don't think we should bother testing all these endpoints because we only need GET ALL and UPDATE ALL for references. And that's all we should test. We can safely remove the rest of the routes and handlers for references. If we need anything else in the future, we can simply implement it.

api/src/components/reference/controller.ts
```diff
@@ -1,6 +1,7 @@
 import * as I from 'interface';
 import * as response from 'lib/response';
 import * as referenceService from 'reference/service';
+import * as publicationService from 'publication/service';
 
 export const get = async (event: I.AuthenticatedAPIRequest<undefined, undefined, I.CreateReferencePath>) => {
```
florin-holhos, line 6: we should call this getAll because it brings all the references for a specific publication

api/src/components/reference/controller.ts
```diff
@@ -14,7 +15,29 @@ export const get = async (event: I.AuthenticatedAPIRequest<undefined, undefined,
 
 export const create = async (event: I.AuthenticatedAPIRequest<I.Reference, undefined, I.CreateReferencePath>) => {
```
florin-holhos, line 16: we can safely remove this handler

api/src/components/reference/controller.ts
```diff
@@ -24,6 +47,28 @@ export const create = async (event: I.AuthenticatedAPIRequest<I.Reference, undef
 
 export const update = async (event: I.AuthenticatedAPIRequest<I.Reference, undefined, I.UpdateReferencePath>) => {
```
florin-holhos, line 48: we can remove this as well

api/src/components/reference/controller.ts
```diff
@@ -45,6 +113,22 @@ export const updateAll = async (event: I.AuthenticatedAPIRequest<I.Reference[],
 
 export const remove = async (event: I.AuthenticatedAPIRequest<undefined, undefined, I.UpdateReferencePath>) => {
```
florin-holhos, line 114: we can remove this one as well

api/src/components/reference/controller.ts
```diff
@@ -53,9 +137,26 @@ export const remove = async (event: I.AuthenticatedAPIRequest<undefined, undefin
     }
 };
 
-export const removeAll = async (event: I.AuthenticatedAPIRequest<undefined, undefined, I.RemoveAllReferencesPath>) => {
+export const removeAll = async (
```
florin-holhos, line 140: we can remove this also, because the updateAll takes care of deleting all the references for a specific publication

api/src/components/reference/service.ts
```diff
@@ -10,7 +10,7 @@ export const getAllByPublication = async (publicationId: string) => {
     return references;
 };
 
-export const create = async (data) => {
+export const create = async (data: I.Reference) => {
```
florin-holhos, line 13: we can remove this

api/src/lib/interface.ts
```diff
@@ -433,8 +433,8 @@ export interface UpdateReferencePath {
     referenceId: string;
 }
 
-export interface RemoveAllReferencesPath {
-    publicationId: string;
+export interface RemoveAllReferencesPathParams {
```
florin-holhos, line 436: we can remove this

api/src/components/reference/__tests__/updateAllReferences.test.ts
shaunajisc: My bad I've changed that 👍

florin-holhos (approved):

PR #254, opened 2022-09-23
OCT-370 Resolve immediate JWT issues

florin-holhos:
The purpose of this PR was to improve user authentication UX and handle JWT token expiry

---

### Acceptance Criteria:

As per [OCT-370](https://jiscdev.atlassian.net/browse/OCT-370)

### Tests:

---

### DIFF
--- a/api/src/components/authorization/controller.ts
+++ b/api/src/components/authorization/controller.ts
@@ -26,8 +26,8 @@ export const authorize = async (event: I.APIRequest<I.AuthorizeRequestBody>): Pr
 
         const userInformation: I.ORCIDUser = ORCIDRequestPublicUserResponse.data;
 
-        const employment = userInformation?.['activities-summary']?.employments['affiliation-group'].map(
-            (employmentItem) => ({
+        const employment =
+            userInformation?.['activities-summary']?.employments['affiliation-group'].map((employmentItem) => ({
                 organisation: employmentItem?.summaries?.[0]?.['employment-summary']?.organization?.name || null,
                 role: employmentItem?.summaries?.[0]?.['employment-summary']?.['role-title'] || null,
                 department: employmentItem?.summaries?.[0]?.['department-name'] || null,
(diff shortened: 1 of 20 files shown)


api/src/components/authorization/controller.ts
```diff
@@ -92,4 +91,9 @@ export const authorize = async (event: I.APIRequest<I.AuthorizeRequestBody>): Pr
         console.log(err);
         return response.json(500, { message: 'Unknown server error.' });
     }
-};
\ No newline at end of file
+};
+
+export const verifyToken = async (event: I.AuthenticatedAPIRequest) => {
```
karmseveeratjisc, line 96: this is really verifying the user or even fetching them, the token has already been processed by the time we get here..

florin-holhos: It's verifying the token and returns a 200 + decoded info it's valid. It's not fetching the user as that would mean to grab it from DB, and that would be too slow for what we need from this endpoint. The decoded info returned is not 100% necessary. 

On the client side, we only care about that 200 response (atm). But on the server side, I thought we could use this endpoint to verify the token + get some basic user info from it like user ID if we need to fetch some data based on that, instead of decoding it server side without knowing 100% if it's valid or not (only the backend can tell that).

See the example below:
image

We could call verifyToken endpoint from "guardPrivateRoute" helper and get the user ID from there if we need it, or to check if the email is verified, before rendering the page.

karmseveeratjisc: all sounds good by my point is the method name suggests the method is verifying the Token when it doesn't, it doesnt verify anything it simply returns the decoded token, perhaps call it fetchDecodedTokenUser or something

florin-holhos: Changed verifyToken to getDecodedUserToken.

ui/src/pages/login.tsx
```diff
@@ -41,39 +34,48 @@ export const getServerSideProps: Types.GetServerSideProps = async (context) => {
             },
             undefined
         );
-        token = response.data;
+        token = response.data as string;
     } catch (err) {
-        const { message } = err as Interfaces.JSONResponseError;
-        error = message;
-        console.log(error);
+        if (process.env.NODE_ENV === 'development') {
```
karmseveeratjisc, line 39: sure there must be a better pattern for this kind of logging but fine for now

karmseveeratjisc (approved):

PR #264, opened 2022-10-07
OCT-405 '& ' breaks search

shaunajisc:
The purpose of this PR was to fix a bug where when the user uses special characters in their search query it breaks the page and returns a 422 error, as per this [Jira ticket](https://jiscdev.atlassian.net/jira/software/projects/OCT/boards/836?selectedIssue=OCT-405). 

This is because the query isn't encoded. I have used `encodeURIComponent()`  to encode the characters but this needs decoding on the backend so the API can interpret the search query.

### DIFF
--- a/api/src/components/user/service.ts
+++ b/api/src/components/user/service.ts
@@ -1,6 +1,7 @@
 import * as client from 'lib/client';
 
 import * as I from 'interface';
+import * as helpers from 'lib/helpers';
 
 export const upsertUser = async (orcid: string, updateUserInformation: I.UpdateUserInformation) => {
     const user = await client.prisma.user.upsert({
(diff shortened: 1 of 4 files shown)


Harry-Jisc:
I'm getting a runtime error on the following characters, what's funny though is on int.octopus these are not causing an issue.

When entering the character "(" without closing it, same with closing bracket, ")", without an opening bracket
Single quotes also causes an error - but not when used within words, only by itself
Colon also causes an error - but not when used within words, only by itself	

They all return the same error below in the API console: 
`"syntax error in tsquery: \"@$(()\"",`

florin-holhos:
> I'm getting a runtime error on the following characters, what's funny though is on int.octopus these are not causing an issue.
> 
> When entering the character "(" without closing it, same with closing bracket, ")", without an opening bracket Single quotes also causes an error - but not when used within words, only by itself Colon also causes an error - but not when used within words, only by itself
> 
> They all return the same error below in the API console: `"syntax error in tsquery: \"@$(()\"",`

I get the same error on int as well

Harry-Jisc (approved):

karmseveeratjisc (approved):

PR #292, opened 2022-11-18
OCT-444 Add capability to add missing references

Harry-Jisc:
The purpose of this PR was to add capability to add missing references.

---

### Acceptance Criteria:
As per [OCT-444](https://jiscdev.atlassian.net/jira/software/projects/OCT/boards/836?selectedIssue=OCT-444)

### DIFF
--- a/api/prisma/seed.ts
+++ b/api/prisma/seed.ts
@@ -18,13 +18,12 @@ export const initialDevSeed = async (): Promise<void> => {
         });
     }
 
-    // Create publications 
+    // Create publications
     for (const publication of SeedData.publicationsDevSeedData) {
         await client.prisma.publication.create({
             // @ts-ignore
             data: publication
         });
-
 
         if (publication.currentStatus === 'LIVE') {
             await client.search.create({
@@ -105,9 +104,8 @@ export const initialDevSeed = async (): Promise<void> => {
     }
 };
 
-
 export const initialProdSeed = async (): Promise<void> => {
-    console.log("running initialProdSeed")
+    console.log('running initialProdSeed');
     // Create users
     await client.prisma.user.createMany({ data: SeedData.usersProdSeedData });
 
@@ -145,32 +143,29 @@ export const initialProdSeed = async (): Promise<void> => {
                     cleanContent: htmlToText.convert(problem.content)
                 }
             });
-        }    
-
+        }
     }
-
 };
 
 switch (process.env.STAGE) {
     case 'prod':
         initialProdSeed()
-        .catch((e) => {
-            console.error(e);
-            // process.exit(1);
-        })
-        .finally(async () => {
-            await client.prisma.$disconnect();
-        });
-    
-    break;
+            .catch((e) => {
+                console.error(e);
+                // process.exit(1);
+            })
+            .finally(async () => {
+                await client.prisma.$disconnect();
+            });
+
+        break;
     default:
         initialDevSeed()
-        .catch((e) => {
-            console.error(e);
-            // process.exit(1);
-        })
-        .finally(async () => {
-            await client.prisma.$disconnect();
-        });
-    
+            .catch((e) => {
+                console.error(e);
+                // process.exit(1);
+            })
+            .finally(async () => {
+                await client.prisma.$disconnect();
+            });
 }
--- a/api/prisma/seed.ts
+++ b/api/prisma/seed.ts
@@ -6,7 +6,7 @@ import * as client from '../src/lib/client';
 
 export const initialDevSeed = async (): Promise<void> => {
     // Create users
-    await client.prisma.user.createMany({ data: SeedData.usersProdSeedData });
+    await client.prisma.user.createMany({ data: SeedData.usersDevSeedData });
 
     const doesIndexExists = await client.search.indices.exists({
         index: 'publications'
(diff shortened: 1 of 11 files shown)


api/prisma/seed.ts
```diff
@@ -173,4 +173,4 @@ switch (process.env.STAGE) {
             await client.prisma.$disconnect();
         });
     
-}
+}
```
florin-holhos, line 176: Please fix all eslint and prettier warnings.
image

api/serverless.yml
```diff
@@ -86,6 +86,13 @@ functions:
                   path: ${self:custom.versions.v1}/publications/{id}
                   method: GET
                   cors: true
+    getPublicationSeed:
+        handler: src/components/publication/routes.getPubSeed
```
florin-holhos, line 90: Given the name, I can't really understand what is this route supposed to do ?!

Harry-Jisc: It gets the publication, for the seed script as it changes what information is required, compared to regular getPublication. I'm not sure what a more fitting name would be tbh 🤔

api/src/components/publication/controller.ts
```diff
@@ -67,6 +67,28 @@ export const get = async (
     }
 };
 
+export const getPubSeed = async (
```
florin-holhos, line 70: It's not clear what is this handler supposed to do by it's name.

Please fix eslint and prettier warnings.

image

api/src/components/publication/controller.ts
```diff
@@ -67,6 +67,28 @@ export const get = async (
...
+    event: I.APIRequest<undefined, I.GetPublicationQueryParams>
+): Promise<I.JSONResponse> => {
+    try {
+
+        const publication = await publicationService.getPubSeed(event.queryStringParameters.title);
```
florin-holhos, line 74: Publications

api/src/components/publication/controller.ts
```diff
@@ -67,6 +67,28 @@ export const get = async (
...
+       
+        if(publication !== null) {
```
florin-holhos: "publication" will never be null because "findMany" always returns an array with the records that match the "where" clause, or an empty array if there's no match. There's no need to check for null here

api/src/components/publication/controller.ts
```diff
@@ -67,6 +67,28 @@ export const get = async (
...
+            return response.json(200, {
+                publication
+            })
+        }
+        return response.json(404, {
```
florin-holhos: I don't a 404 is suitable here. Because it's not like fetching /publications/{publicationId} where you're trying to get a specific publication.
You can either change the service from "findMany" to "findFirst" and then you can keep this 404 as it is. Either simply return whatever the service is returning => array with pubs or empty array

api/prisma/seed.ts
```diff
@@ -145,32 +143,29 @@ export const initialProdSeed = async (): Promise<void> => {
...
                 }
             });
-        }    
-
+        }
     }
-
 };
 
 switch (process.env.STAGE) {
```
florin-holhos, line 150: This switch should be wrapped in a function

florin-holhos: Actually, ignore this atm because this file needs to be refactored anyways.
There's a good example on prisma docs on how this file should look like: https://www.prisma.io/docs/guides/database/seed-database

karmseveeratjisc: This file wasn't changed I think except @Harry-Jisc  ran prettier on it

api/serverless.yml
florin-holhos: Maybe something like "getSeedPublications" would be easier to understand

api/serverless.yml
```diff
@@ -86,6 +86,13 @@ functions:
...
+        events:
+            - http:
+                path:  ${self:custom.versions.v1}/publication/seed
```
florin-holhos, line 93: Again, the naming feels a bit awkward. Maybe something like "/seed-data-publications" would be suitable for this

api/src/components/publication/controller.ts
```diff
@@ -67,6 +67,20 @@ export const get = async (
     }
 };
 
+export const getPubSeed = async (
```
florin-holhos, line 70: you can replace "getPubSeed" with "getSeedDataPublications"

api/src/lib/interface.ts
```diff
@@ -101,8 +101,8 @@ export interface GetPublicationPathParams {
     id: string;
 }
 
-export interface GetLinksQueryParams {
-    publicationID: string
+export interface GetPublicationQueryParams {
```
florin-holhos, line 104: Maybe "GetSeedDataPublicationsFilters" would be more suitable name for this. It's still long, but at least it's specific.
You can also add the "userId" here so we can see exactly all the filters applied to the query. Because atm the userId is hardcoded in the service of this route and that's not ideal.

Finally, your route would look something like this: /seed-data-publications?title={publicationTitle}&userId={userId}

api/src/components/publication/routes.ts
```diff
@@ -15,6 +15,11 @@ export const get = middy(publicationController.get)
     .use(middleware.httpJsonBodyParser())
     .use(middleware.authentication(true));
 
+export const getPubSeed = middy(publicationController.getPubSeed)
```
florin-holhos, line 18: getSeedDataPublications

api/src/components/publication/routes.ts
```diff
@@ -15,6 +15,11 @@ export const get = middy(publicationController.get)
...
+    .use(middleware.doNotWaitForEmptyEventLoop({ runOnError: true, runOnBefore: true, runOnAfter: true }))
+    .use(middleware.httpJsonBodyParser())
+    .use(middleware.authentication(true));
```
florin-holhos, line 21: you can remove authentication middleware from this route because it is not used

api/src/components/publication/service.ts
```diff
@@ -192,6 +192,30 @@ export const get = async (id: string) => {
     return publication;
 };
 
+export const getPubSeed = async (title: string) => {
```
florin-holhos, line 195: "getSeedDataPublications" or something similar

api/src/components/publication/service.ts
```diff
@@ -192,6 +192,30 @@ export const get = async (id: string) => {
...
+    const publication = await client.prisma.publication.findMany({
```
florin-holhos, line 196: please use plural for arrays "publication" => "publications"

florin-holhos (approved):

api/src/lib/interface.ts
Harry-Jisc: I've changed the name of the interface to the one suggested, but no worried about the userId, it's only ever octopus that will have seed data, so don't need to adapt for other users.

florin-holhos (approved):

PR #294, opened 2022-11-24
OCT-403 Science Octopus account returns 502

florin-holhos:
The purpose of this PR was to add pagination for /users/{userId}/publications in order to avoid loading all user publications at once which will break the app for some specific users like Science Octopus.

Removed user publications from /users/{userId} response

Removed user object from /users/{userId}/publications

User and user publications can now be fetched independently without breaking the app.

Moved e2e tests in project's root folder.

---

### Acceptance Criteria:

---

### Tests:

Updated getUserPublication tests in order to match the changes that were made.

image

Created an ORCID account for e2e tests. (will provide the credentials in private)
Added e2e tests for author profile.
Updated login helper to cope with cases like: ORCID "Authorize access" and New user email verification
Fixed all failing tests (18)
Fixed all flaky tests. (some of the tests were randomly failing and they were really hard to debug)

image

---


Harry-Jisc (dismissed):

api/src/components/user/controller.ts
```diff
@@ -32,20 +32,23 @@ export const get = async (event: I.OptionalAuthenticatedAPIRequest<undefined, un
...
 
-        const statuses: Array<I.ValidStatuses> = isAccountOwner ? ['DRAFT', 'LIVE'] : ['LIVE'];
-
-        const userPublications = await userService.getPublications(event.pathParameters.id, statuses, isAccountOwner);
+        const user = isAccountOwner ? event.user : await userService.get(event.pathParameters.id);
 
-        if (!userPublications) {
-            return response.json(404, { message: 'User not found' });
+        if (!user) {
+            return response.json(400, { message: 'Invalid user id' });
```
karmseveeratjisc, line 43: I think 404 was correct  for this case, 400 would be correct if the 'user' part of the path had to match a particular pattern , but we currently allow anything here

ui/src/pages/authors/[id]/index.tsx
```diff
@@ -1,22 +1,29 @@
-import React from 'react';
+import React, { useMemo, useState } from 'react';
```
Comment (karmseveeratjisc) [inline ui/src/pages/authors/[id]/index.tsx:1]: this feels like the e2e test should be updated, it's a very simple test at the moment and I know updating it won't be simple, is this functionality 'core' enough though to justify writing the test?

api/src/components/user/controller.ts
florin-holhos: 404 is not correct here, because we're not querying the user itself, it's more of a client error: 400 bad request, because the client should know exactly the id of the user it's trying to get the publications for, it's not like guessing.

karmseveeratjisc: that is not what 400 is for, you should be able to determine if you need to return a 400 by purely examining the request and not by e.g calling a service to look up a user

karmseveeratjisc: If a user is deleted their id is still 'valid' - if someone has bookmarked the user's profile page that should then return 404  (or 410) not 400, and if someone types arandom user id  that does not make it a bad request (and the client has no way of knowing it is not a 'real' user)

karmseveeratjisc: https://github.com/paralleldrive/cuid userids are cuids so we _could_ put validation in that the the userid has to start 'c' but that's about it

ui/src/pages/authors/[id]/index.tsx
Comment (florin-holhos) [reply ui/src/pages/authors/[id]/index.tsx:1]: We don't currently have any meaningful e2e tests for this page and there are some errors when running them.
image

Should I add e2e tests for this page in this PR?

Comment (karmseveeratjisc) [reply ui/src/pages/authors/[id]/index.tsx:1]: 'no meaningful tests' suggests yes - if they won't take long , the existing ones were passing as few weeks ago...

api/src/components/user/controller.ts
karmseveeratjisc: we'll leave it as 400 but maybe change the message to 'User not found'...

ui/src/pages/authors/[id]/index.tsx
Comment (florin-holhos) [reply ui/src/pages/authors/[id]/index.tsx:1]: added/fixed e2e tests

api/src/components/user/controller.ts
florin-holhos: changed the message back to 'User not found'

PR #295, opened 2022-11-29
OCT-444 Fixes raised from int seed data ingest. 

Harry-Jisc:
The purpose of this PR was fix errors that has been discovered during testing of seed ingest on int.octopus.ac

### DIFF
--- a/seed-data-scripts/seed/seed.js
+++ b/seed-data-scripts/seed/seed.js
@@ -338,7 +338,7 @@ const formatReference = (references, publicationId) => {
       },
       paragraphsArray[i]
     );
-    let referenceText = newReference.text.replace(/^""+$/g, "");
+    let referenceText = newReference.text.replace(/"(.*?)"/g, "$1")
     newReference = { ...newReference, text: referenceText };
     referencesArray.push(newReference);
   }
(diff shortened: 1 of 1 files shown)


seed-data-scripts/seed/seed.js
```diff
@@ -338,7 +338,8 @@ const formatReference = (references, publicationId) => {
       },
       paragraphsArray[i]
     );
-
+    let referenceText = newReference.text.replace(/"(.*?)"/g, "$1")
```
karmseveeratjisc, line 341: not sure what this is trying to match?
for
"beginning ""middle"" end" 

it will give 
beginning middle end 

we want:

beginning "middle" end

karmseveeratjisc: actually seems to give:

beginning middle" end

karmseveeratjisc: nope right first time, all " gone

karmseveeratjisc (approved):

seed-data-scripts/seed/seed.js
karmseveeratjisc: working now

PR #311, opened 2022-12-20
OCT-466 Separate author and publication search

gbJisc:
The purpose of this PR was...

- Separate the Author and Publication search into two separate pages under /search/<searchType>
- Apply new redirect to search pages dropdown based on selection (requested)
- Apply redirect to search/publications when path is /search (requested default)
- Remove superfluous code from each component/props based on search requirements/filters
- Update typescript generics and constraints based on the above
- Correct api.search() calls across project with optional publication param & type constraint
- Keep existing api call path to users/?search (not authors/?search)

### Acceptance Criteria:

---

### Tests:

---


ui/src/pages/search/authors/index.tsx
```diff
@@ -0,0 +1,315 @@
...
+                                        details={[
+                                            'Try a different search criteria.',
+                                            'If you think something is wrong, please contact the helpdesk.'
+                                        ]}
+                                    />
+                                )}
+
+                                {!isValidating && results?.data?.length && (
+                                    <>
+                                        <div className="rounded">
```
florin-holhos, line 244: Search results are not aligned properly. They should extend to 12 columns for now. Until we decide if and what filters to add for authors, on the left hand side, like we have for publications.
image

ui/next.config.js
```diff
@@ -1,6 +1,16 @@
 /** @type {import('next').NextConfig} */
 const nextConfig = {
-    reactStrictMode: true
+    reactStrictMode: true,
+
+    async redirects() {
```
florin-holhos, line 5: This is now showing a blank page for 1-2 seconds when pressing the Search button from the Header.
Not sure if using a middleware for this redirect has the same issue. But for now, we can make that Search button from the Header to redirect the user straight to '/search/publications' instead of '/search' and avoid the redirect when clicking on it.

ui/src/pages/search/authors/index.tsx
```diff
@@ -0,0 +1,315 @@
...
+                <section className="container mx-auto px-8 py-8 lg:gap-4 lg:pt-16 lg:pb-0">
+                    <Components.PageTitle text={`Search results ${query ? `for ${query}` : ''}`} />
+                </section>
+                <section
+                    id="content"
+                    className="container mx-auto grid grid-cols-1 px-8 lg:grid-cols-12 lg:gap-x-16 lg:gap-y-8"
+                >
+                    <fieldset className="col-span-12 mb-8 grid w-full grid-cols-12 items-end gap-y-4 gap-x-6 lg:mb-0 lg:gap-x-6 2xl:gap-x-10">
+                        <legend className="sr-only">Search options</legend>
+
```
florin-holhos, line 140: We should extract this search markup and logic in a new component and use that component in both: /publications and /authors

florin-holhos (approved):

PR #312, opened 2022-12-21
merge main to int

karmseveeratjisc:
The purpose of this PR was...

---

### Acceptance Criteria:

---

### Tests:

---

### DIFF
--- a/api/.env.example
+++ b/api/.env.example
@@ -30,3 +30,6 @@ DOI_PREFIX=10.82259
 DATACITE_ENDPOINT=https://api.test.datacite.org/dois
 DATACITE_USER=user
 DATACITE_PASSWORD=password
+
+MAIL_SERVER=localhost
+LOCALSTACK_SERVER=localhost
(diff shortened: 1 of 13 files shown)


api/.env.example
```diff
@@ -30,3 +30,6 @@ DOI_PREFIX=10.82259
 DATACITE_ENDPOINT=https://api.test.datacite.org/dois
 DATACITE_USER=user
 DATACITE_PASSWORD=password
+
+MAIL_SERVER=localhost
```
florin-holhos, line 34: MAIL_SERVER needs to be 0.0.0.0 on localhost. Email verification doesn't work with MAIL_SERVER=localhost

e2e/tests/helpers.ts
```diff
@@ -46,7 +46,7 @@ export const login = async (page: Page) => {
     const newPage = await context.newPage();
 
     // navigate to MailHog and take the last verification code
-    await newPage.goto("http://localhost:8025/");
+    await newPage.goto(`http://${process.env.MAIL_SERVER}:8025/`);
```
florin-holhos, line 49: This change breaks the login in e2e tests, because MAIL_SERVER variable cannot be 'localhost', it needs to be 0.0.0.0 in order to be able to send verification emails locally. 

After updating MAIL_SERVER to 0.0.0.0 the verification email is sent successfully but this url becomes 'http://0.0.0.0:8025' and the email verification fails in e2e tests.

PR #321, opened 2023-01-05
OCT-497 Validate emails

Harry-Jisc:
The purpose of this PR was ensure emails were validated before attempting to add/send them.

---

### Acceptance Criteria:

As per [OCT-497](https://jiscdev.atlassian.net/browse/OCT-497?atlOrigin=eyJpIjoiMTk4YWUzMzMzYTc2NDkxYmEzNDE0ZjVmNTZhM2JlNDYiLCJwIjoiaiJ9)

[OCT-497]: https://jiscdev.atlassian.net/browse/OCT-497?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ui/src/components/Publication/Creation/CoAuthor.tsx
```diff
@@ -15,9 +16,25 @@ const CoAuthor: React.FC = (): React.ReactElement => {
 
     const [loading, setLoading] = React.useState(false);
     const [coAuthor, setCoAuthor] = React.useState('');
+    const [emailValidated, setEmailValidated] = React.useState<Boolean>(true);
```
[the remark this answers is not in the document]
Harry-Jisc: I'll remove just to keep the consistency with the others

ui/src/components/Publication/Creation/CoAuthor.tsx
```diff
@@ -15,9 +16,25 @@ const CoAuthor: React.FC = (): React.ReactElement => {
...
 
+    const handleEmailChange = (event: React.ChangeEvent<HTMLInputElement>) => {
```
gbJisc, line 21: (Minor) moving forward it would be good to introduce consistency with return types (here `: void `) when adding new functions/methods.

ui/src/lib/helpers.tsx
```diff
@@ -337,3 +337,9 @@ export const getFullDOIsStrings = (text: string) =>
 export const getDOIsFromText = (text: string) => text.match(/(10\.[0-9a-zA-Z]+\/(?:(?!["&\'])\S)+)\b/g) || [];
 
 export const validateDOI = (value: string) => /(10\.[0-9a-zA-Z]+\/(?:(?!["&\'])\S)+)\b/.test(value);
+
+export const validateEmail = (email: string): Boolean => {
+    const regex = /[a-zA-Z0-9'_\\-\\.\\+]+@[a-zA-Z0-9'_\\-\\.]+[.][a-zA-Z0-9'_\\-\\.]+/;
+    if (email.match(regex)) return true;
```
florin-holhos, line 343: I know email validation it's a vast topic, but this validation accepts inputs like "test test test@test.com" (see the screenshot below) and then, the backend returns an error (which we're not displaying) and the user has no clue what happened. 

I think we also need to display any error messages coming back from the backend in the same way we are displaying the ones for frontend validation.

The regex needs to capture the entire input value, that means it needs to capture the start "^" and the end "$". Currently, this validation just tells if the string contains a potential email address. You can use "regex.test(email)" instead of "email.match(regex)" in this case because "match" returns an array with all matches it finds, and you don't need that array, you just need to know if it's valid or not.

image

ui/src/pages/verify.tsx
```diff
@@ -45,9 +45,24 @@ const Verify: Types.NextPage<Props> = (props): React.ReactElement => {
     const [loading, setLoading] = React.useState(false);
     const [error, setError] = React.useState('');
     const [success, setSuccess] = React.useState(false);
+    const [emailValidated, setEmailValidated] = React.useState<Boolean>(true);
```
karmseveeratjisc, line 48: <Boolean> not needed here either

ui/src/lib/helpers.tsx
```diff
@@ -338,4 +338,10 @@ export const getDOIsFromText = (text: string) => text.match(/(10\.[0-9a-zA-Z]+\/
 
 export const validateDOI = (value: string) => /(10\.[0-9a-zA-Z]+\/(?:(?!["&\'])\S)+)\b/.test(value);
 
+export const validateEmail = (email: string): Boolean => {
```
karmseveeratjisc, line 341: might be useful to have some actual tests to test some valid and invalid emails

karmseveeratjisc: could be a unit test for this method (though we don't have anything like that at the moment but we probably should) or indirectly via a test for the CoAuthor component

karmseveeratjisc (changes requested):
we need tests

ui/src/__tests__/lib/Helpers.test.tsx
```diff
@@ -0,0 +1,20 @@
+import * as Helpers from '@helpers';
+
+describe('Validate emails tests', () => {
+
+    it('Valid emails', () => {
+        const emails = ["email@example.com", "firstname.lastname@example.com", "email@subdomain.example.com", "email@example-one.com", "email@example.name", "email@example.museum", "email@example.co.jp", "firstname-lastname@example.com"]
```
karmseveeratjisc, line 6: maybe add a '+' example here too:

email+octopus@gmail.com

ui/src/lib/helpers.tsx
```diff
@@ -338,4 +338,11 @@ export const getDOIsFromText = (text: string) => text.match(/(10\.[0-9a-zA-Z]+\/
...
+    // const regex = /^[a-zA-Z][\w\.-]*[a-zA-Z0-9]@[a-zA-Z][\w\.-]*[a-zA-Z0-9]\.[a-zA-Z][a-zA-Z\.]*[a-zA-Z]$/;
```
gbJisc, line 342: can we rm commented line?

ui/src/lib/helpers.tsx
```diff
@@ -338,4 +338,11 @@ export const getDOIsFromText = (text: string) => text.match(/(10\.[0-9a-zA-Z]+\/
...
+    const regex = /^([\w+-]+\.)*[\w+-]+@[a-zA-Z][\w\.-]*[a-zA-Z0-9]\.[a-zA-Z][a-zA-Z\.]*[a-zA-Z]$/;
+    if (regex.test(email)) return true;
+    return false;
```
gbJisc, line 345: neater would be:

`return (regex.test(email));`

karmseveeratjisc (approved):

PR #337, opened 2023-02-07
OCT-533 Standardise eslint rules, remove prettier, introduce automation

gbJisc:
The purpose of this PR was...
Standardise formatting (api) via prettier, disable eslint formatting
Automate prettier --write and eslint --fix on pre-commit hook
Manually fix any outstanding issues from api path

### Acceptance Criteria:
(api path only)
git hook is added on npm install
eslint --fix runs on pre-commit on staged files if < 10 or all api files if > 10 staged when committing branch
any formatting issues are auto fixed
any errors that cannot be auto-fixed are communicated to user
commit fails if errors present
commit succeeds if no errors/only warnings
prettier and eslint are run on entire repo by github action to check nothing has managed to bypass hook

### Tests:

---

### DIFF
--- a/api/package.json
+++ b/api/package.json
@@ -23,7 +23,8 @@
         "reindex": "ts-node prisma/reindex.ts",
         "format:check": "npx prettier --check src/",
         "format:write": "npx prettier --write src/",
-        "lint": "eslint src/"
+        "lint:check": "eslint src/",
+        "lint:fix": "eslint --fix src/"
     },
     "dependencies": {
         "@aws-sdk/client-ses": "^3.76.0",
--- a/api/package.json
+++ b/api/package.json
@@ -16,6 +16,7 @@
         "seed": "ts-node prisma/seed.ts"
     },
     "scripts": {
+        "prepare": "cd .. && husky install",
         "start:local": "nodemon --config nodemon.json",
         "seed:local": "STAGE=local npx prisma migrate reset --force",
         "test:local": "STAGE=local jest --runInBand --testTimeout=60000",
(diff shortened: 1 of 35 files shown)


karmseveeratjisc (commented):
think prettier.json can be removed from api folder now?

api/package.json
```diff
@@ -23,7 +23,8 @@
         "reindex": "ts-node prisma/reindex.ts",
         "format:check": "npx prettier --check src/",
         "format:write": "npx prettier --write src/",
-        "lint": "eslint src/"
+        "lint:check": "eslint src/",
```
karmseveeratjisc, line 26: we should probably run this in github actions now?

api/package.json
```diff
@@ -23,7 +23,8 @@
         "reindex": "ts-node prisma/reindex.ts",
         "format:check": "npx prettier --check src/",
```
karmseveeratjisc, line 24: these can go? prettier has been removed?

karmseveeratjisc (changes requested):

api/package.json
karmseveeratjisc: we've decided to keep it and only do formatting using it, eslint just does linting

api/package.json
karmseveeratjisc: hmm the checks dont make sense if force running prettier and linting (as the linting 'fixes' prettier), either we assume  prettier and linting have run (due to pre-commit hook) or we can only run one of them (lint check perhaps better than nothing? then if someone has managed to push unprettiered/unlinted code this will fail?)

karmseveeratjisc: done

karmseveeratjisc:
> think prettier.json can be removed from api folder now?

no we are keeping prettier now

karmseveeratjisc (approved):

florin-holhos (approved):

PR #341, opened 2023-02-16
OCT-529 - Indicate to authors whether action is needed from their /author page

Harry-Jisc:
The purpose of this PR was to Indicate to authors whether action is needed from their /author page

---

### Acceptance Criteria:
As Per [OCT-529](https://jiscdev.atlassian.net/jira/software/projects/OCT/boards/836?selectedIssue=OCT-529)

[OCT-529]: https://jiscdev.atlassian.net/browse/OCT-529?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/e2e/tests/LoggedIn/publish.e2e.spec.ts
+++ b/e2e/tests/LoggedIn/publish.e2e.spec.ts
@@ -659,7 +659,7 @@ const confirmCoAuthorInvitation = async (browser: Browser, user: Helpers.TestUse
     // clicking 'I am an author' link is blocked by cors
     const invitationLink = await page2
         .frameLocator('iframe')
-        .locator("a.button:has-text('I am an author')")
+        .locator("a:has-text('I am an author')")
         .getAttribute('href');
 
     // navigate to that link instead
@@ -738,6 +738,9 @@ export const verifyPublicationIsDisplayedAsDraftForCoAuthor = async (
     await page.locator(PageModel.header.myProfileButton).click();
 
     await expect(page.locator(PageModel.myProfile.draftPublicationHeader)).toHaveText('Draft publications');
+
+    // Confirm publication states under review
+    await expect(page.locator(`a:has-text("${publicationTitle}")`)).toContainText('Under Review');
 
     // // Confirm publication is showed as draft
     await page.locator(`a:has-text("${publicationTitle}")`).click();
@@ -1254,7 +1257,7 @@ test.describe('Publication flow + co-authors', () => {
         await page.close();
     });
 
-    test('Coauthored publications show on your own profile', async ({ browser }) => {
+    test('Coauthored publications show on your own profile with correct publication status', async ({ browser }) => {
         const context = await browser.newContext();
         const page = await context.newPage();
         await page.goto(Helpers.UI_BASE);
@@ -1304,8 +1307,15 @@ test.describe('Publication flow + co-authors', () => {
         await page.waitForSelector(PageModel.publish.publishButton);
         await expect(page.locator(PageModel.publish.publishButton)).toBeEnabled();
 
+        // verify the status is set to 'ready to publish' for this publication
+        await page.locator(PageModel.header.usernameButton).click();
+        await page.locator(PageModel.header.myProfileButton).click();
+        await expect(page.locator(`a:has-text("${publicationTitle}")`)).toContainText('Ready to publish');
+        // go back to publication
+        await page.locator(`a:has-text("${publicationTitle}")`).click();
+
         // publish the new publication
-        page.locator(PageModel.publish.publishButton).click();
+        await page.locator(PageModel.publish.publishButton).click();
         await Promise.all([page.waitForNavigation(), page.locator(PageModel.publish.confirmPublishButton).click()]);
 
         // veryify publication is displayed as live on co-author profile
--- a/ui/src/components/Publication/SimpleResult/index.tsx
+++ b/ui/src/components/Publication/SimpleResult/index.tsx
@@ -9,63 +9,76 @@ type Props = {
     user: Interfaces.User;
 };
 
-const SimpleResult: React.FC<Props> = (props): React.ReactElement => (
-    <div className="w-full rounded border border-transparent bg-white-50 p-3 shadow transition-colors duration-500 dark:border-teal-500 dark:bg-transparent dark:shadow-none">
-        <div className="flex justify-between gap-2">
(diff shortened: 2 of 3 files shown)


ui/src/components/Publication/SimpleResult/index.tsx
```diff
@@ -9,48 +9,79 @@ type Props = {
...
+                                : 'text-purple-500 dark:text-purple-300'
+                        } flex items-center gap-2 font-semibold leading-3 transition-colors duration-500`}
+                    >
+                        {props.publication.currentStatus === 'LIVE' ? (
+                            <OutlineIcons.ExternalLinkIcon className="inline h-4 w-4 leading-3 text-grey-600 dark:text-teal-500" />
+                        ) : (
+                            <OutlineIcons.PencilAltIcon className="inline h-4 w-4 leading-3 text-grey-600 dark:text-teal-500" />
+                        )}
+
+                        {React.useMemo(
```
florin-holhos, line 55: Please remove 'useMemo' from the jsx. This hook is used to compute and memoize a value BEFORE the render method ('return' statement in this case). It also has some missing deps

ui/src/components/Publication/SimpleResult/index.tsx
```diff
@@ -9,48 +9,79 @@ type Props = {
...
-                    className={`${
-                        props.publication.currentStatus === 'LIVE'
-                            ? 'text-teal-500 dark:text-teal-300'
-                            : 'text-purple-500 dark:text-purple-300'
-                    } flex items-center gap-2 font-semibold leading-3 transition-colors duration-500`}
-                >
-                    {props.publication.currentStatus === 'LIVE' ? (
-                        <OutlineIcons.ExternalLinkIcon className="inline h-4 w-4 leading-3 text-grey-600 dark:text-teal-500" />
+const SimpleResult: React.FC<Props> = (props): React.ReactElement => {
+    const publicationStatus = (publication: Interfaces.UserPublication, user: Interfaces.User) => {
```
florin-holhos, line 13: This can either be a function outside the component (because currently this function is re-created every time when this component re-renders) and just call it inside the component, or, it can be a useMemo inside the component which computes the status that needs to be shown inside the jsx

ui/src/components/Publication/SimpleResult/index.tsx
```diff
@@ -9,48 +9,79 @@ type Props = {
...
+        if (publication.createdBy === user.id) {
+            if (publication.coAuthors.length === 1) return 'Draft';
+
+            if (publication.coAuthors.every((coAuthors) => coAuthors.confirmedCoAuthor === true))
+                return 'Ready to publish';
+
+            return 'Draft';
+        }
+
+        const status = publication.coAuthors.map((coAuthor) => {
```
florin-holhos, line 25: the status needs to be a single string, not an array

ui/src/components/Publication/SimpleResult/index.tsx
```diff
@@ -9,48 +9,79 @@ type Props = {
...
+            if (coAuthor.linkedUser === props.user.id) {
```
florin-holhos, line 26: please don't interpolate component props with function arguments. The 'user' argument is the same as 'props.user'

e2e/tests/LoggedIn/publish.e2e.spec.ts
```diff
@@ -659,7 +659,7 @@ const confirmCoAuthorInvitation = async (browser: Browser, user: Helpers.TestUse
     // clicking 'I am an author' link is blocked by cors
     const invitationLink = await page2
         .frameLocator('iframe')
-        .locator("a.button:has-text('I am an author')")
+        .locator("a:has-text('I am an author')")
```
florin-holhos, line 662: this has already been updated on the main branch, please always get the latest code from main

e2e/tests/LoggedIn/publish.e2e.spec.ts
```diff
@@ -1304,8 +1307,15 @@ test.describe('Publication flow + co-authors', () => {
...
+        // verify the status is set to 'ready to publish' for this publication
+        await page.locator(PageModel.header.usernameButton).click();
+        await page.locator(PageModel.header.myProfileButton).click();
+        await expect(page.locator(`a:has-text("${publicationTitle}")`)).toContainText('Ready to publish');
+        // go back to publication
+        await page.locator(`a:has-text("${publicationTitle}")`).click();
+
         // publish the new publication
-        page.locator(PageModel.publish.publishButton).click();
+        await page.locator(PageModel.publish.publishButton).click();
```
florin-holhos, line 1318: this is also updated on the main branch

ui/src/components/Publication/SimpleResult/index.tsx
```diff
@@ -9,48 +9,80 @@ type Props = {
...
+    const publicationStatus = (publication: Interfaces.UserPublication, user: Interfaces.User) => {
+        if (publication.currentStatus === 'LIVE') return 'Live';
+
+        if (publication.createdBy === user.id) {
+            if (publication.coAuthors.length === 1) return 'Draft';
+
+            if (publication.coAuthors.every((coAuthors) => coAuthors.confirmedCoAuthor === true))
+                return 'Ready to publish';
+
+            return 'Draft';
```
florin-holhos, line 22: I think this should return 'Under Review' instead of 'Draft'. Please check with the PO

ui/src/components/Publication/SimpleResult/index.tsx
```diff
@@ -9,48 +9,80 @@ type Props = {
...
+        }
+
+        let status = '';
+        publication.coAuthors.forEach((coAuthor) => {
+            if (coAuthor.linkedUser === user.id) {
+                if (coAuthor.confirmedCoAuthor) {
+                    status = 'Under Review';
+                    return;
```
florin-holhos, line 30: These returns are not stopping the iterations when you find what you need, the forEach will still loop through each element till the end of the array even if you don't need that. For these types of checks there are better array methods like: find, some and findIndex. You can also use a simple for loop and a break inside, but usually the array methods mentioned above are the way to go.

ui/src/components/Publication/SimpleResult/index.tsx
Harry-Jisc: Clarified with Tim, we are keeping as is for now, as how the locking works will change this in [OCT-521](https://jiscdev.atlassian.net/browse/OCT-521) and the ticket is being updated.

florin-holhos (approved):

PR #343, opened 2023-02-21
OCT-521 - Locking publications whilst approval is pending

Harry-Jisc:
The purpose of this PR was to lock publications for editing while under review by other authors.


api/src/components/coauthor/controller.ts
```diff
@@ -253,13 +253,19 @@ export const updateConfirmation = async (
             });
         }
 
-        // Is the publication in draft?
+        // Is the publication in locked mode?
```
florin-holhos, line 256: here, we just need to check if currentStatus !== 'LOCKED', therefore confirmation cannot be  processed

api/src/components/coauthor/controller.ts
```diff
@@ -253,13 +253,19 @@ export const updateConfirmation = async (
...
         if (publication.currentStatus === 'LIVE') {
             return response.json(403, {
                 message: 'This publication is LIVE and therefore cannot be edited.'
```
florin-holhos, line 259: This message needs to be changed because the confirmation cannot be processed while the currentStatus is LIVE or DRAFT

api/src/components/coauthor/controller.ts
```diff
@@ -253,13 +253,19 @@ export const updateConfirmation = async (
...
             });
         }
 
+        if (publication.currentStatus === 'DRAFT') {
```
florin-holhos, line 263: This if is not needed. We only need to know if the currentStatus !== 'LOCKED', and return the error message based on the current status: LIVE or DRAFT

florin-holhos: minor eslint issue
image

api/src/components/coauthor/service.ts
```diff
@@ -142,10 +142,11 @@ export const updateConfirmation = async (publicationId: string, userId: string,
     return updateCoAuthor;
 };
 
-export const resetCoAuthors = async (publicationId: string) => {
+export const resetCoAuthors = async (publicationId: string, correspondingAuthor: string) => {
```
florin-holhos, line 145: this service shouldn't ask for the creator, it can process everything by the publicationId

api/src/components/link/controller.ts
```diff
@@ -74,9 +74,11 @@ export const deleteLink = async (event: I.AuthenticatedAPIRequest<undefined, und
 
         if (
             link.publicationFromRef.currentStatus !== 'DRAFT' ||
-            !link.publicationFromRef.publicationStatus.every((status) => status.status === 'DRAFT')
+            !link.publicationFromRef.publicationStatus.every((status) => status.status !== 'LIVE')
```
florin-holhos, line 77: is this check really needed? since no publication can go back from LIVE to DRAFT.

currentStatus !== 'DRAFT' should be enough

api/src/components/coauthor/service.ts
Harry-Jisc: It needs to corresponding author, otherwise set their automatic approval to false.

api/src/components/link/controller.ts
Harry-Jisc: Might be needed, might not be, always been in, since there is mention of this 'HIDDEN' status type, in which LIVE publications can go from LIVE -> HIDDEN -> LIVE, not sure if that is still planned or not

api/src/lib/interface.ts
```diff
@@ -265,7 +265,7 @@ export interface DeleteLinkPathParams {
     id: string;
 }
 
-export type ValidStatuses = 'DRAFT' | 'LIVE';
+export type ValidStatuses = 'DRAFT' | 'LIVE' | 'LOCKED';
```
florin-holhos, line 268: These statuses should match the ones in the DB. If we don't use HIDDEN, then we should remove it from schema.prisma also. We have more mismatches like this that we will need to address

api/src/components/coauthor/service.ts
florin-holhos: I know, but we can get the corresponding author by the publicationId

api/src/components/coauthor/controller.ts
```diff
@@ -253,10 +253,13 @@ export const updateConfirmation = async (
...
-        // Is the publication in draft?
-        if (publication.currentStatus === 'LIVE') {
+        // Is the publication in locked mode?
+        if (publication.currentStatus !== 'LOCKED') {
             return response.json(403, {
-                message: 'This publication is LIVE and therefore cannot be edited.'
+                message:
+                    publication.currentStatus === 'LIVE'
+                        ? 'This publication is LIVE and therefore cannot be edited.'
+                        : 'This publication is in DRAFT and therefore cannot be edited.'
```
florin-holhos, line 262: Error messages are not related to what this controller is actually doing, the old error message was wrong also. Through this controller we're not editing the publication.

If currentStatus is DRAFT we can say something like: 'This publication is not ready for approval yet'
If it's LIVE, we can say something like: 'You cannot approve a LIVE publication'

api/src/components/publication/controller.ts
```diff
@@ -235,8 +236,22 @@ export const updateStatus = async (
...
-        if (publication?.currentStatus !== 'DRAFT') {
-            return response.json(404, { message: 'A status of a publication that is not in DRAFT cannot be changed.' });
+        if (publication?.currentStatus === 'LIVE') {
+            return response.json(403, {
+                message: 'A status of a publication that is not in DRAFT or LOCKED cannot be changed.'
+            });
+        }
+
+        if (publication?.currentStatus === 'LOCKED') {
+            // Update status to 'DRAFT'
```
florin-holhos, line 246: This is not updating the status just to DRAFT, corresponding author can also update status to LIVE bypassing co-authors approvals...

We need to check the status provided by the user input also and if they are allowed to change the status from one to another

DRAFT -> LIVE
DRAFT -> LOCKED -> LIVE
DRAFT -> LOCKED -> DRAFT
LIVE -> nowhere...

ui/src/components/Publication/SimpleResult/index.tsx
```diff
@@ -16,10 +16,12 @@ const SimpleResult: React.FC<Props> = (props): React.ReactElement => {
         if (publication.createdBy === user.id) {
             if (publication.coAuthors.length === 1) return 'Draft';
 
+            if (publication.currentStatus === 'DRAFT') return 'Draft';
```
florin-holhos, line 19: the first two conditions can be combined

ui/src/layouts/BuildPublication.tsx
```diff
@@ -123,19 +123,16 @@ const BuildPublication: React.FC<BuildPublicationProps> = (props) => {
         []
     );
 
-    const CheckCoAuthorsToApprove = useCallback(
+    const publicationHasCoAuthors = useCallback(
```
florin-holhos, line 126: Is this function really needed? it is only used in one useMemo that returns a boolean.

ui/src/layouts/BuildPublication.tsx
```diff
@@ -304,7 +303,7 @@ const BuildPublication: React.FC<BuildPublicationProps> = (props) => {
         [checkRequiredApproval, isReadyToPreview, store]
     );
 
-    const coAuthorsToApprove = useMemo(() => CheckCoAuthorsToApprove(store).ready, [CheckCoAuthorsToApprove, store]);
+    const hasCoAuthors = useMemo(() => publicationHasCoAuthors(store).ready, [publicationHasCoAuthors, store]);
```
florin-holhos, line 306: not really sure if we need a useMemo just to check if the publication has coAuthors

ui/src/pages/publications/[id]/index.tsx
```diff
@@ -234,6 +243,31 @@ const Publication: Types.NextPage<Props> = (props): React.ReactElement => {
         }
     }, [confirmation, publicationData?.id, router]);
 
+    const handleUnlock = async () => {
+        const confirmed = await confirmation(
+            'Unlocking this publication for editing will invalidate all existing author approvals. Authors will be invited to approve your new changes once you have finished editing',
+            'Are you sure you want to cancel and unlock?',
+            <OutlineIcons.CloudUploadIcon className="h-10 w-10 text-grey-600" aria-hidden="true" />,
```
Comment (florin-holhos) [inline ui/src/pages/publications/[id]/index.tsx:250]: I don't think 'CloudUploadIcon' matches the confirmation modal. This needs to be checked with PO

ui/src/layouts/BuildPublication.tsx
```diff
@@ -304,7 +291,7 @@ const BuildPublication: React.FC<BuildPublicationProps> = (props) => {
         [checkRequiredApproval, isReadyToPreview, store]
     );
 
-    const coAuthorsToApprove = useMemo(() => CheckCoAuthorsToApprove(store).ready, [CheckCoAuthorsToApprove, store]);
+    const hasCoAuthors = store.coAuthors.length > 1 ? true : false;
```
florin-holhos, line 294: why not simply "const hasCoAuthors = store.coAuthors.length > 1" ?

florin-holhos (approved):

ui/src/layouts/BuildPublication.tsx
karmseveeratjisc: why indeed 👍

PR #353, opened 2023-03-10
OCT-521 - Fix request approval

Harry-Jisc:
The purpose of this PR was to fix request approval for older publications that don't have the corresponding author listed.

### DIFF
--- a/ui/src/layouts/BuildPublication.tsx
+++ b/ui/src/layouts/BuildPublication.tsx
@@ -291,7 +291,9 @@ const BuildPublication: React.FC<BuildPublicationProps> = (props) => {
         [checkRequiredApproval, isReadyToPreview, store]
     );
 
-    const hasCoAuthors = store.coAuthors.length > 1 ? true : false;
+    const hasCoAuthors = !store.coAuthors
+        .filter((coAuthor) => coAuthor.linkedUser !== props.publication.createdBy)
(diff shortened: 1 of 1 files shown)


ui/src/layouts/BuildPublication.tsx
```diff
@@ -291,7 +291,9 @@ const BuildPublication: React.FC<BuildPublicationProps> = (props) => {
         [checkRequiredApproval, isReadyToPreview, store]
     );
 
-    const hasCoAuthors = store.coAuthors.length > 1 ? true : false;
+    const hasCoAuthors = !store.coAuthors
```
florin-holhos, line 294: Please pay attention to naming. The name needs to match the checking, otherwise it can be very confusing and can lead to bugs if other devs will use this boolean just by it's name. You're checking if not all the co-authors have been invited, therefore the name: 'hasUninvitedCoAuthors'.

ui/src/layouts/BuildPublication.tsx
```diff
@@ -291,7 +291,9 @@ const BuildPublication: React.FC<BuildPublicationProps> = (props) => {
...
+        .filter((coAuthor) => coAuthor.linkedUser !== props.publication.createdBy)
```
florin-holhos, line 295: 'filter' is not needed, you can do the check without it

florin-holhos (approved):

PR #354, opened 2023-03-14
OCT-507 - Confirming affiliations

Harry-Jisc:
The purpose of this PR was to force user to declare their affiliation status and prevent them from publishing if they do not.

---

### Acceptance Criteria:

As per [OCT-507](https://jiscdev.atlassian.net/browse/OCT-507?atlOrigin=eyJpIjoiMTFjNDViZDE4YTdhNGNjNDhhYTA5ZGZjNTJmM2ExMGUiLCJwIjoiaiJ9)

[OCT-507]: https://jiscdev.atlassian.net/browse/OCT-507?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/prisma/schema.prisma
```diff
@@ -86,6 +86,8 @@ model Publication {
     funders                            Funders[]
     fundersStatement                   String?
     affiliations                       Affiliations[]
+    confirmNoAffliations               Boolean?
+    affiliationsStatus                 Boolean?               @default(true)
```
florin-holhos, line 90: Since this is holding a boolean, I think 'hasAffiliations' would have been more appropriate.
Status usually refers to a string like: ACTIVE, INACTIVE, LOCKED, LIVE, DRAFT and so on...

api/src/components/affiliations/controller.ts
```diff
@@ -52,10 +52,9 @@ export const destroy = async (
             });
         }
 
-        //check that the publication is live
-        if (publication.currentStatus !== 'DRAFT') {
+        if (publication.currentStatus === 'LIVE') {
```
florin-holhos, line 55: The previous check was alright. This is now allowing corresponding user to remove affiliations while the publication is LOCKED

api/src/components/affiliations/controller.ts
```diff
@@ -74,3 +73,36 @@ export const destroy = async (
...
+        const publication = await publicationService.get(event.pathParameters.id);
+
+        //check that the publication exists
+        if (!publication) {
+            return response.json(404, {
+                message: 'This publication does not exist.'
+            });
+        }
+
+        if (publication.currentStatus === 'LIVE') {
```
florin-holhos, line 90: The same as above. This is allowing corresponding user to  remove affiliations while the publication is LOCKED...

ui/src/components/Publication/Creation/KeyInformation.tsx
```diff
@@ -13,6 +13,23 @@ import * as Types from '@types';
 const KeyInformation: React.FC = (): React.ReactElement => {
     const title = Stores.usePublicationCreationStore((state) => state.title);
     const updateTitle = Stores.usePublicationCreationStore((state) => state.updateTitle);
+
+    const affiliationsStatus = Stores.usePublicationCreationStore(
```
florin-holhos, line 17: All these lines could be simplified by using destructuring assignment. Please update the old assignments like this as well (within this file). This looks like a lot of code for so little, just to grab a bunch of properties from the store.

ui/src/components/Publication/Creation/KeyInformation.tsx
```diff
@@ -107,21 +124,54 @@ const KeyInformation: React.FC = (): React.ReactElement => {
...
+                                Does this publication have any affiliations?
                             </span>
-                            <Components.Button
-                                title="Search for your organisation's ROR"
-                                href="https://ror.org/"
-                                openNew
-                                endIcon={
-                                    <OutlineIcons.SearchIcon className="h-6 w-6 text-teal-500 transition-colors duration-500 dark:text-white-50" />
-                                }
+                            <Components.RadioConfirm
```
florin-holhos, line 135: There's an issue with the props provided to this component
image

I see 'status' prop is undefined in some cases. That prop should always be a boolean => undefined means false

florin-holhos: UI state is not updating correctly.
Steps to replicate the issue:
1. Select 'Yes' and add an affiliation
image
2. Select 'No', click confirm checkbox and save the publication
3. Select 'Yes' again. The affiliation is deleted from DB but it still shows in the UI because the state was not updated
image
4. Save the publication (DELETE affiliations endpoint is called again, not sure why?!)
5. Refresh the page - affiliations won't show up anymore

ui/src/components/RORForm/index.tsx
```diff
@@ -232,6 +233,11 @@ const RORForm: React.FC<FormProps> = (props): React.ReactElement => {
             props.type == 'funders'
                 ? updateFunders([...funders, createdRorRecord])
                 : updateAffiliations([...affiliations, createdRorRecord]);
+
+            if (props.type === 'affiliations') {
+                updateConfirmNoAffliations(false);
```
florin-holhos, line 238: Is there a reason why updating this boolean when adding an affiliation and not when the user clicks the 'Yes' radio button? 

User can click 'Yes' and save the publication without adding any affiliations and this flag will remain true. Then if the user clicks 'No' again, the confirmation checkbox is selected by default, because this flag remained true all this time.

ui/src/components/RadioConfirm/index.tsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import * as Stores from '@stores';
+
+type Props = {
+    status: boolean | undefined;
```
florin-holhos, line 5: This prop must be a boolean, always. Otherwise the radio input will switch from uncontrolled to controlled input and it will misbehave. There's a warning into the console about this.

ui/src/layouts/BuildPublication.tsx
```diff
@@ -60,9 +60,29 @@ const BuildPublication: React.FC<BuildPublicationProps> = (props) => {
                     message: 'You have selected there is a conflict of interest, please provide a reason.'
                 };
             }
-            if (typeof store.conflictOfInterestStatus == 'undefined') {
+
+            if (typeof store.conflictOfInterestStatus === 'undefined') {
```
florin-holhos, line 64: Looking at the requirements, this option should never be undefined. It's 'true' by default, that means the only other option is 'false'. But never undefined

ui/src/stores/publicationCreation.ts
```diff
@@ -122,6 +124,10 @@ let store: any = (set: (params: any) => void) => ({
     // Affiliations
     affiliations: [],
     updateAffiliations: (affiliations: Interfaces.Affiliations[]) => set(() => ({ affiliations })),
+    confirmNoAffliations: undefined,
```
florin-holhos, line 127: this should be false by default, not undefined

ui/src/stores/publicationCreation.ts
```diff
@@ -122,6 +124,10 @@ let store: any = (set: (params: any) => void) => ({
...
+    updateConfirmNoAffliations: (confirmNoAffliations: boolean) => set(() => ({ confirmNoAffliations })),
+    affiliationsStatus: undefined,
```
florin-holhos, line 129: this should be true by default, not undefined

ui/src/layouts/BuildPublication.tsx
Harry-Jisc: This is not in the scope of the ticket, this is to do with conflict of interest, I just updated the equality to check with an extra '='

api/serverless.yml
```diff
@@ -448,3 +448,10 @@ functions:
                   path: ${self:custom.versions.v1}/publications/{id}/affiliation/{affiliation}
                   method: DELETE
                   cors: true
+    deleteAllAffiliations:
+        handler: src/components/affiliations/routes.destroyAll
+        events:
+            - http:
+                  path: ${self:custom.versions.v1}/publications/{id}/affiliations/
```
florin-holhos, line 455: just a minor thing, but I guess we don't need the forward slash at the end

api/prisma/migrations/20230313145217_add_confirm_affiliations/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "Publication" ADD COLUMN     "confirmNoAffliations" BOOLEAN DEFAULT FALSE;
```
florin-holhos, line 2: *Affiliations

florin-holhos: You can install this VSCode extension that will help you with typos https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker

api/prisma/migrations/20230309162534_add_affiliations_status/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "Publication" ADD COLUMN     "affiliationsStatus" BOOLEAN DEFAULT TRUE;
```
florin-holhos, line 2: Please provide better names for these two columns, a boolean should start with is, has, should, and so on... Using these names to check something will make the code hard to read/understand.

ui/src/components/Publication/Creation/KeyInformation.tsx
```diff
@@ -107,21 +132,69 @@ const KeyInformation: React.FC = (): React.ReactElement => {
...
+                                        required
+                                        id="confirm"
+                                        name="confirm"
+                                        type="checkbox"
+                                        checked={confirmNoAffliations}
+                                        disabled={affiliationsStatus}
+                                        onChange={() => updateConfirmNoAffliations(!confirmNoAffliations)}
+                                        className="rounded-sm border-teal-500 bg-white-50 outline-0 transition-colors duration-500 focus:ring-2 focus:ring-yellow-400 disabled:opacity-50"
+                                    />
+                                    <span className="dark:text-grey-200' ml-2 block text-grey-500 text-grey-800 transition-colors duration-500 dark:text-white-50">
```
florin-holhos, line 192: image

ui/src/components/Publication/Creation/KeyInformation.tsx
```diff
@@ -6,18 +6,43 @@ import * as Components from '@components';
...
+        affiliationsStatus,
+        affiliations,
+        updateAffiliations,
+        updateAffiliationsStatus,
+        confirmNoAffliations,
+        updateConfirmNoAffliations,
+        licence,
+        updateLicence
+    } = Stores.usePublicationCreationStore(
+        (state: Types.PublicationCreationStoreType) => ({
```
florin-holhos, line 32: this selector function is not needed

api/prisma/migrations/20230309162534_add_affiliations_status/migration.sql
florin-holhos: I'm sorry, but i don't agree with this. If something was poorly named by others, that doesn't mean we need to continue in the same manner just to be "consistent". Being consistent with wrong stuff it's not a good thing.

ui/src/components/RadioConfirm/index.tsx
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import * as Stores from '@stores';
+
+type Props = {
+    status: boolean;
+    updateStatusTrue: () => void;
```
florin-holhos, line 6: a single 'onChange' handler provided via props it's more than enough, you can do different stuff based on the parameter true/false in the parent component.

ui/src/layouts/BuildPublication.tsx
```diff
@@ -60,9 +60,29 @@ const BuildPublication: React.FC<BuildPublicationProps> = (props) => {
...
+            }
+
+            if (!store.affiliationsStatus && !store.hasNoAffiliations) {
+                ready = {
+                    ready: false,
+                    message: 'You have selected there are affiliations, please provide them'
+                };
+            }
+
+            if (typeof store.affiliationsStatus === 'undefined') {
```
florin-holhos, line 82: this code block is obsolete. 'store.affiliationsStatus' will never be undefined with the current codebase

ui/src/layouts/BuildPublication.tsx
```diff
@@ -60,9 +60,29 @@ const BuildPublication: React.FC<BuildPublicationProps> = (props) => {
                     message: 'You have selected there is a conflict of interest, please provide a reason.'
                 };
             }
-            if (typeof store.conflictOfInterestStatus == 'undefined') {
+
+            if (typeof store.conflictOfInterestStatus === 'undefined') {
                 ready = { ready: false, message: 'You must select a conflict of interest option' };
             }
+
+            if (store.affiliationsStatus && !store.affiliations.length) {
```
florin-holhos, line 68: you also need to add these validations on the backend as well, because currently the user can publish even if they selected 'Yes' but didn't add any affiliations at all. The opposite is true as well, they can publish even if they selected 'No' but they didn't confirm they have no affiliations.

api/src/components/affiliations/controller.ts
```diff
@@ -74,3 +73,36 @@ export const destroy = async (
...
+
+        if (event.user.id !== publication.user.id) {
+            return response.json(403, {
+                message: 'You do not have permissions to delete an affiliation for this publication.'
+            });
+        }
+
+        await affiliationService.destroyAll(event.pathParameters.id);
+
+        return response.json(200, { message: `Affliations removed for publication ${event.pathParameters.id}.` });
```
florin-holhos, line 104: just a typo on `Affliations removed...

florin-holhos (approved):

PR #358, opened 2023-03-20
OCT-547 - Pass given & family names over to datacite

Harry-Jisc:
The purpose of this PR was to ensure datacite presents the authors names properly by passing given & family names over to datacite for new and existing publications.

---

### Acceptance Criteria:
As Per [OCT-547](https://jiscdev.atlassian.net/browse/OCT-547?atlOrigin=eyJpIjoiNGM4NzgyODdmMjU1NGQxNThhYjk1MWUzMjIzY2MwNDYiLCJwIjoiaiJ9)

[OCT-547]: https://jiscdev.atlassian.net/browse/OCT-547?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/src/scripts/updateDoiNames.ts
```diff
@@ -0,0 +1,165 @@
+import * as client from '../lib/client';
+import * as helpers from '../lib/helpers';
+
+const updateDoiNames = async (): Promise<void> => {
+    const publications = await client.prisma.publication.findMany({
```
karmseveeratjisc, line 5: feels a little like should already be a method to get pubs and there are, but those are in API code, guess this is a 'one shot' script but might set a pattern that is not ideal..

Harry-Jisc: there is, as you say, but its an API call and required  some parameters for open search, it was just easier to do this, a bit more "raw" but it is, only a one shot script, I don't think it will set a pattern, specially if we don't keep it around afterwards?

karmseveeratjisc: PR to delete then 😉

karmseveeratjisc (approved):

PR #366, opened 2023-04-03
OCT-595 - Make guidance more obvious

Harry-Jisc:
The purpose of this PR was to make the guidance more obvious.

---

### Acceptance Criteria:

As per [OCT-595](https://jiscdev.atlassian.net/browse/OCT-595?atlOrigin=eyJpIjoiNjM0ODE3N2VmMjk4NGVkYzk5YmZkNTViM2NmMzk3MmQiLCJwIjoiaiJ9)

Comment on ticket:

> I have spoken to Tim about the positioning of the “Author Guide” button - This will no longer be to the right to the search, instead the order will be: Learn More - Author Guide - Search

[OCT-595]: https://jiscdev.atlassian.net/browse/OCT-595?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ui/src/components/Nav/index.tsx
+++ b/ui/src/components/Nav/index.tsx
@@ -18,6 +18,24 @@ const Nav: React.FC = (): React.ReactElement => {
                 value: Config.urls.browsePublications.path
             },
             {
+                label: 'How To',
+                value: '',
+                subItems: [
+                    {
+                        label: 'FAQ',
+                        value: Config.urls.faq.path
+                    },
+                    {
+                        label: 'Author Guide',
+                        value: Config.urls.authorGuide.path
+                    },
+                    {
+                        label: 'Our Aims',
+                        value: Config.urls.octopusAims.path
+                    }
+                ]
+            },
+            {
                 label: 'Publish',
                 value: Config.urls.createPublication.path
             }
@@ -44,7 +62,7 @@ const Nav: React.FC = (): React.ReactElement => {
         return menuItems;
     }, [user]);
 
-    return isDesktop ? <Components.NavDesktop items={items} /> : <Components.NavMobile items={items} />;
+    return isDesktop ? <Components.NavDesktop user={user} items={items} /> : <Components.NavMobile items={items} />;
 };
 
 export default Nav;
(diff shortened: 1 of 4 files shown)


ui/src/components/Nav/index.tsx
```diff
@@ -44,7 +62,7 @@ const Nav: React.FC = (): React.ReactElement => {
         return menuItems;
     }, [user]);
 
-    return isDesktop ? <Components.NavDesktop items={items} /> : <Components.NavMobile items={items} />;
+    return isDesktop ? <Components.NavDesktop user={user} items={items} /> : <Components.NavMobile items={items} />;
```
florin-holhos, line 65: user prop it's not needed after fixing the data-testid mentioned above

ui/src/pages/index.tsx
```diff
@@ -78,6 +78,14 @@ const Home: Types.NextPage<Props> = (props): React.ReactElement => {
                                     Learn more
                                 </span>
                             </Components.Link>
+                            <Components.Link
```
florin-holhos, line 81: elements need to wrap at some point
image

florin-holhos: this is still broken
image

buttons and search should be aligned in some way, maybe something similar to this
image

ui/src/pages/index.tsx
```diff
@@ -69,15 +69,23 @@ const Home: Types.NextPage<Props> = (props): React.ReactElement => {
                             upon the latest work.
                         </p>
 
-                        <div className="mx-auto flex w-fit space-x-6">
+                        <div className="mx-auto flex w-fit flex-wrap justify-around gap-y-5 space-x-6">
```
florin-holhos, line 72: instead of gap-y and space-x, you can use just 'gap'

ui/src/pages/index.tsx
florin-holhos: it still looks misaligned, and the gap between 'Learn more' and 'Author Guide' is too big. The buttons also wrap too early.
image
image

Harry-Jisc: Maybe, but "too big" is subjective, I personally think that this design looks nicer. I'm not really sure how to resolve this, other than asking Tim, since it's just opinion over opinion, then again that would be another opinion, as long as it functions without "breaking" anything, like it did in the original, I'd say is acceptable ?

54565856f3175887a0bc424aa39af989

c1d7eaeca2a251312d133c8954727a66

florin-holhos (approved):

PR #367, opened 2023-04-03
Suggestion: Remove z-index for username on desktop nav.

AshleyRedman:
On desktop, when opening the command palette, the display of the username shows above the cp overlay. The issue looks to be desktop only.

Screenshot 2023-04-03 at 20 54 08

### DIFF
--- a/ui/src/components/Nav/Desktop/index.tsx
+++ b/ui/src/components/Nav/Desktop/index.tsx
@@ -15,7 +15,7 @@ const Desktop: React.FC<Props> = (props): React.ReactElement => (
             {props.items.map((item) => (
                 <li key={item.value} className="first:ml-0 last:mr-0 md:mx-1 lg:mx-2">
                     {item.subItems?.length ? (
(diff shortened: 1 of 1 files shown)


ui/src/components/Nav/Desktop/index.tsx
```diff
@@ -15,7 +15,7 @@ const Desktop: React.FC<Props> = (props): React.ReactElement => (
             {props.items.map((item) => (
                 <li key={item.value} className="first:ml-0 last:mr-0 md:mx-1 lg:mx-2">
                     {item.subItems?.length ? (
-                        <HeadlessUI.Menu as="div" className="relative z-50 inline-block text-left">
+                        <HeadlessUI.Menu as="div" className="relative inline-block text-left">
```
florin-holhos, line 18: this can be part of [OCT-550](https://jiscdev.atlassian.net/browse/OCT-550)

shaunajisc:
This is no longer relavent as we have removed the quick search

PR #370, opened 2023-04-06
OCT-588 API Request approval not checking completion

florin-holhos:
The purpose of this PR was to prevent publications from being LOCKED without filling out the required fields like title, main text, co-authors list and so on...

---

### Acceptance Criteria:

As per [OCT-588](https://jiscdev.atlassian.net/browse/OCT-588)

---

### Tests:

---

### Screenshots:

[OCT-588]: https://jiscdev.atlassian.net/browse/OCT-588?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/prisma/seeds/publications.ts
+++ b/api/prisma/seeds/publications.ts
@@ -94,6 +94,47 @@ const publicationSeeds = [
         }
     },
     {
+        id: 'locked-publication-problem-confirmed-co-authors',
+        title: 'LOCKED Publication PROBLEM confirmed co-authors',
+        type: 'PROBLEM',
+        licence: 'CC_BY',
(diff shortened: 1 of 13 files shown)


Harry-Jisc:
Ideally this should be under the /status/{status} endpoint and including locked in there. While this does fix the bug and the ticket might not specifically say to add it under there, this is a good opportunity to fix something that is wrong with the system at the moment, albeit initiated by myself, we shouldn't continue to do it this way if it bad API design.

Harry-Jisc (approved):
Tested locally and everything works as did before, and the bug has been fixed. Tests also have been added!

PR #379, opened 2023-04-17
OCT-610 - Updated .env.examples

Harry-Jisc:
The purpose of this PR was to update the .env.examples.

---

### Acceptance Criteria:

As Per [OCT-610](https://jiscdev.atlassian.net/browse/OCT-610?atlOrigin=eyJpIjoiZGFhY2E0YzI1NTU5NGExMDg3MzA1YjFmMTU2MjE4NjgiLCJwIjoiaiJ9)

[OCT-610]: https://jiscdev.atlassian.net/browse/OCT-610?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


shaunajisc (approved):
Looks good! 👍 I think we may need to update/ remove the .env part in the [api read me](https://github.com/JiscSD/octopus/tree/main/api) too? Just something I noticed when setting up, although I think we could just remove the code block there and reference the .env example file.

Harry-Jisc:
> Looks good! 👍 I think we may need to update/ remove the .env part in the [api read me](https://github.com/JiscSD/octopus/tree/main/api) too? Just something I noticed when setting up, although I think we could just remove the code block there and reference the .env example file.

Yes, good spot! I'll do it as part of this. will request re-approval when done and remove it, be better to just point to the .env.example rather than maintain it in two places!

shaunajisc (approved):
Nice one! 👍🏻

api/.env.example
```diff
@@ -31,5 +29,5 @@ DATACITE_ENDPOINT=https://api.test.datacite.org/dois
 DATACITE_USER=user
 DATACITE_PASSWORD=password
 
-MAIL_SERVER=0.0.0.0
-LOCALSTACK_SERVER=localhost
\ No newline at end of file
+MAIL_SERVER=localhost
```
florin-holhos, line 32: Is there a reason for changing MAIL_SERVER? This is breaking the mailing service for me

PR #383, opened 2023-05-05
OCT-596 Corresponding author specific affiliations

florin-holhos:
The purpose of this PR was to provide a way for authors to add their ORCID affiliations. 

- added two new endpoints: GET /orcid-affiliations and PUT /publications/{id}/my-affiliations
- backend allows each author to add his/her ORCID affiliations
- added new column in the CoAuthors table called 'affiliations' which is of type 'Json[]' where we can store a mapped version of the affiliations coming from ORCID
- updated data that's being sent to data cite when publishing. Each creator now has his own list of affiliations. Contributors only contain the creator of the publication

---

### Acceptance Criteria:

As per [OCT-596](https://jiscdev.atlassian.net/browse/OCT-596) + ticket comments

---

### Tests:

Updated backend and UI tests

---

### Screenshots:

image

image

[OCT-596]: https://jiscdev.atlassian.net/browse/OCT-596?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ui/src/pages/publications/[id]/flag/[flagId]/index.tsx
```diff
@@ -82,12 +80,11 @@ type Props = {
...
-        props.flagUrl,
-        {
-            fallback: props.fallback
-        }
-    );
+    const { data, isValidating, error, mutate } = useSWR<Interfaces.FlagWithComments>(props.flagUrl, {
+        fallback: props.fallback
+    });
+
+    console.log(data);
```
Comment (finlay-jisc) [inline ui/src/pages/publications/[id]/flag/[flagId]/index.tsx:87]: This console log can probably be removed

Comment (florin-holhos) [reply ui/src/pages/publications/[id]/flag/[flagId]/index.tsx]: yep, removed.

finlay-jisc (approved):
Looks good to me but should have another approver too.

shaunajisc (dismissed):
Nice one! I've tested locally and it's working as expected ☺️ 👍🏻 

There's just one small bug - if a user hasn’t specified their affiliations it prevents them from saving, this is inconsistent with other required fields in the publication (eg. Main text). Screenshot below

Also, RE this error message, I think we should change the error message to something like: 

> Please fill out your affiliation information  

Because the error message as it stands:

> Please add your affiliations if your research was not conducted independently

may confuse users who are independent but haven’t ticked the box.


https://user-images.githubusercontent.com/100135505/236475567-d696bae1-54ce-4b91-b53a-622fbcdde6a6.mov

florin-holhos:
> Nice one! I've tested locally and it's working as expected ☺️ 👍🏻
> 
> There's just one small bug - if a user hasn’t specified their affiliations it prevents them from saving, this is inconsistent with other required fields in the publication (eg. Main text). Screenshot below
> 
> Also, RE this error message, I think we should change the error message to something like:
> 
> > Please fill out your affiliation information
> 
> Because the error message as it stands:
> 
> > Please add your affiliations if your research was not conducted independently
> 
> may confuse users who are independent but haven’t ticked the box.
> 
>  Screen.Recording.2023-05-05.at.13.02.02.mov

1. updated the way we check author affiliations.
2. updated response error message to 'Please fill out your affiliation information.'

finlay-jisc (approved):
I can see that the bug Shauna found is fixed and that the error message has changed* so happy to approve.

* I wasn't able to surface the error message in the UI anymore (looks like the isReadyToLock check happens first and produces its own message), but it is checked in the API tests so I think it's OK.

PR #386, opened 2023-05-11
OCT-614 - Queued generation of PDFs 

Harry-Jisc:
The purpose of this PR was to create the code which allows for PDFs to be queued. Currently this will create a local queue, but it is not processed when a message is sent. PDF generation will only happen when this has been deployed, since we have had issues with a local versions offline-sqs.

---

### Acceptance Criteria:

As Per [OCT-614](https://jiscdev.atlassian.net/browse/OCT-614?atlOrigin=eyJpIjoiMTI3YTNlNzg0OWQ3NGU2ZWE5NmE4NzFhYjA4OGY2MjIiLCJwIjoiaiJ9)

[OCT-614]: https://jiscdev.atlassian.net/browse/OCT-614?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/src/components/sqs/handler.ts
```diff
@@ -0,0 +1,18 @@
+import axios from 'axios';
+
+export const generatePDFs = async (event): Promise<object> => {
+    const promises = event.Records.forEach(async (record) => {
+        const { body } = record;
+        console.log('Publication ID:', body);
```
finlay-jisc, line 6: We should remove this log unless we need it for debugging etc.

finlay-jisc (commented):
Looks fine to me, but think another dev ought to formally approve.

api/prisma/seed.ts
```diff
@@ -1,6 +1,6 @@
 import htmlToText from 'html-to-text';
 import s3 from '../src/lib/s3';
-
+import { createQueue, getQueue } from '../src/lib/sqs';
```
florin-holhos, line 3: import * as sqs from '../src/lib/sqs';

or

import sqs from '../src/lib/sqs';

Please check coding conventions on the readme file https://github.com/JiscSD/octopus#team-code-conventions

api/src/components/sqs/handler.ts
```diff
@@ -0,0 +1,17 @@
+import axios from 'axios';
+
+export const generatePDFs = async (event): Promise<object> => {
+    const promises = event.Records.forEach(async (record) => {
+        const { body } = record;
+
+        try {
+            await axios.get(`https://${process.env.STAGE}.api.octopus.ac/v1/publications/${body}/pdf`);
+        } catch (err) {
```
florin-holhos, line 9: async await doesn't work inside forEach loop. This catch block won't catch any axios errors because all the http requests are fired immediately and not awaiting for the actual response.

florin-holhos (changes requested):

api/src/lib/sqs.ts
```diff
@@ -0,0 +1,50 @@
+import AWS from 'aws-sdk';
+
+const queueUrl =
+    process.env.STAGE === 'local'
+        ? 'http://localhost:4566/000000000000/science-octopus-pdf-queue-local'
+        : `https://sqs.eu-west-1.amazonaws.com/948306873545/science-octopus-pdf-queue-${process.env.STAGE}`;
+
+const config = {
+    region: 'eu-west-1',
+    endpoint: process.env.STAGE === 'local' ? process.env.LOCALSTACK_SERVER : 'https://sqs.eu-west-1.amazonaws.com'
```
florin-holhos, line 10: As I mentioned above, checking the STAGE wouldn't be necessary if we have an SQS_ENDPOINT env var

api/prisma/seed.ts
```diff
@@ -110,6 +110,14 @@ export const initialDevSeed = async (): Promise<void> => {
                 .promise();
             console.log('Bucket created');
         }
+
+        try {
+            await sqs.getQueue('science-octopus-pdf-queue-local');
```
florin-holhos, line 115: The queue name is going to be 'science-octopus-pdf-queue-local' for all the stages?

Harry-Jisc: This is for local development

api/src/components/sqs/handler.ts
```diff
@@ -0,0 +1,15 @@
...
+            console.log(err);
+        }
+    }
+
+    return {};
```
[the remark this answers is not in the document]
Harry-Jisc: So no, technically, but eslint complains if I don't specify a return type and If I do, I need to return something.

api/prisma/seed.ts
florin-holhos: Well, since you create the queue with a prefix + stage [here](https://github.com/JiscSD/octopus/pull/386/files#:~:text=QueueName%3A%20%60science%2Doctopus%2Dpdf%2Dqueue%2D%24%7Bprocess.env.STAGE%7D%60%2C)

Then you should get that queue back by using the same prefix + stage.
Hardcoding this will also break things if we run 'seed' command on other environments than local. (like int for example)

Harry-Jisc: We won't wont this on int, we have different seen data for that environment, plus this is out of scope, the local development has been pushed into another ticket for investigation and currently this works.

api/src/lib/sqs.ts
```diff
@@ -0,0 +1,46 @@
+import AWS from 'aws-sdk';
+
+const queueUrl = process.env.SQS_ENDPOINT!;
```
florin-holhos, line 3: ESlint shows a warning on this line. I guess removing that non-null assertion will produce other TS issues. A workaround for this would be to check if that 'queueUrl' is undefined and throw an error if so (since you want to make sure that SQS_ENDPOINT env var is available), like we do on ui and e2e for env vars.

api/prisma/seed.ts
florin-holhos: There's no guarantee we will never run this script on int or any other environments. Why not simply updating the 'getQueue' method since it's bad practice to hardcode strings like that? 'createQueue' was perfectly fine as it was.

Harry-Jisc: It's largely irrelevant what whether it's hardcoded or not, since even if we did run it on another environment, the seed script would error as it needs the docker containers running, and if we did that we would use the local env vars anyway? no? but for the sake of getting this approved, I will change it.

api/src/lib/helpers.ts
```diff
@@ -914,3 +914,11 @@ export const createPublicationFooterTemplate = (publication: I.Publication): str
 };
 
 export const isEmptyContent = (content: string): boolean => (content ? /^(<p>\s*<\/p>)+$/.test(content) : true);
+
+export const checkEnvVariable = (variable: string | undefined): string => {
+    if (variable === undefined) {
+        throw new Error(`Environment Variable ${variable} is undefined.`);
```
florin-holhos, line 920: if the env variable is undefined the error message will say: 
_Environment Variable undefined is undefined._

Harry-Jisc: I get this when the variable is undefined Environment Variable 'Queue URL' is undefined.?

florin-holhos (approved):

florin-holhos (approved):

PR #400, opened 2023-05-23
OCT-642 - Able to request approval without entering affiliations

finlay-jisc:
The purpose of this PR was to fix a bug where co-author approval can be requested when:

- no conflict of interest status is selected
- no affiliations have been entered & the corresponding author has not said they're independent

There were some commits that swapped the order to locking the publication before sending out approval emails, but Florin clarified that this is a mistake so I have undone those changes.

---

### Acceptance Criteria:

In this case:
- The button should be disabled.
- If an attempt is made to request approval, the checks for locking the publication should happen before emails are sent.

---

### Tests:

UI
Screenshot 2023-05-23 160407
API
Screenshot 2023-05-23 160823

---

### DIFF
--- a/api/src/components/coauthor/service.ts
+++ b/api/src/components/coauthor/service.ts
@@ -171,7 +171,6 @@ export const resetCoAuthors = async (publicationId: string) => {
         },
         data: {
             confirmedCoAuthor: false,
-            approvalRequested: false,
             code: cuid()
         }
     });
--- a/api/src/components/coauthor/service.ts
+++ b/api/src/components/coauthor/service.ts
@@ -171,6 +171,7 @@ export const resetCoAuthors = async (publicationId: string) => {
         },
         data: {
             confirmedCoAuthor: false,
+            approvalRequested: false,
             code: cuid()
         }
     });
--- a/api/src/components/coauthor/service.ts
+++ b/api/src/components/coauthor/service.ts
@@ -158,12 +158,20 @@ export const resetCoAuthors = async (publicationId: string) => {
     const resetCoAuthors = await client.prisma.coAuthors.updateMany({
         where: {
             publicationId,
-            NOT: {
-                linkedUser: publication?.createdBy
-            }
+            OR: [
+                {
+                    NOT: {
+                        linkedUser: publication?.createdBy
+                    }
+                },
+                {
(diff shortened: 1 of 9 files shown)


api/src/components/coauthor/service.ts
```diff
@@ -158,12 +158,20 @@ export const resetCoAuthors = async (publicationId: string) => {
...
+                    }
+                },
+                {
+                    linkedUser: null
+                }
+            ]
         },
         data: {
             confirmedCoAuthor: false,
+            approvalRequested: false,
```
florin-holhos, line 174: this change triggers the initial co-author invitation email every time the corresponding author unlocks the publication for changes and then request approvals again. 

As a co-author, if I already approved the publication, my approval should be dismissed and I should get a notification about "changes have been made...", but i'm getting the initial invitation email as well
image

finlay-jisc: That's true. Just removed this line.

api/src/components/coauthor/service.ts
```diff
@@ -158,9 +158,16 @@ export const resetCoAuthors = async (publicationId: string) => {
...
-                linkedUser: publication?.createdBy
-            }
+            OR: [
+                {
+                    NOT: {
+                        linkedUser: publication?.createdBy
+                    }
+                },
+                {
+                    linkedUser: null
```
florin-holhos, line 168: This makes the invitation code (which is embedded into "Confirm & Review Publication" link ) reset every time the corresponding user unlocks the publications. The co-author cannot confirm or approve anymore because his initial invitation code is not valid anymore.

Steps to reproduce:

1. log in as a corresponding author and create a new publication
2. add a co-author
3. fill out everything and hit 'Request Approval' button
4. unlock the publication for changes (at this stage the invitation code for co-author is being reset)
5. hit 'Request Approval' button again
6. open a new browser instance and navigate to mailhog
7. try to accept the invitation that has been sent to co-author

image

finlay-jisc: Ah, I can see now why it was that way. At the time I thought this block was supposed to be catching null values too but I didn't realise the impact! I've reversed it.

florin-holhos (approved):

PR #407, opened 2023-06-05
Set up @octopus.ac S3 bucket

shaunajisc:
The purpose of this PR was to create an s3 bucket to store inbound emails to a number of @octopus.ac email addresses, which will then be forwarded on using a Lambda function in a future ticket.

---

### Acceptance Criteria:

As per [OCT-660](https://jiscdev.atlassian.net/jira/software/projects/OCT/boards/836?selectedIssue=OCT-660)

[OCT-660]: https://jiscdev.atlassian.net/browse/OCT-660?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


infra/modules/s3/emails.tf
```diff
@@ -0,0 +1,30 @@
+provider "aws" {
+  region = "eu-west-1"
+}
+
+resource "aws_s3_bucket_acl" "my_bucket" {
+  bucket = "octopus-email-bucket"
```
florin-holhos, line 6: Into the ticket description and also on aws, this bucket has a different name. I know TF it's going to create it with whatever name we give it but it might be confusing while working with this bucket

infra/modules/s3/emails.tf
```diff
@@ -0,0 +1,30 @@
+provider "aws" {
+  region = "eu-west-1"
+}
+
+resource "aws_s3_bucket_acl" "my_bucket" {
```
florin-holhos, line 5: Instead of 'my_bucket' we can be more specific. I think something like 'email_forwarding_bucket' would be clearer

infra/modules/s3/emails.tf
```diff
@@ -0,0 +1,30 @@
+provider "aws" {
```
Harry-Jisc, line 1: You don't need to have the provider set here, this is handled in provider.tf. 

These resources can be included in s3/main.tf instead of being separated into emails.tf. Otherwise you'll need to reference them from s3/main.tf anyway.

infra/modules/s3/emails.tf
shaunajisc: Good point thanks - I'll change it to `email_forwarding_bucket`

infra/modules/s3/emails.tf
shaunajisc: Ah ok thanks - that makes sense

Harry-Jisc (dismissed):

Harry-Jisc (approved):

PR #411, opened 2023-06-08
OCT-678 -  Investigate Alternatives to crypto-random-string

Harry-Jisc:
The purpose of this PR was to investigate alternatives to crypto-random-string. 

The new codes are still 7 characters long and only use uppercase characters and number.

---
### NPM Package

Last published 4 months ago
https://www.npmjs.com/package/otp-generator

No current vulnerabilities specified on sneak for version 4.0.1:
https://snyk.io/advisor/npm-package/otp-generator

---

### Acceptance Criteria:
As Per [OCT-678](https://jiscdev.atlassian.net/browse/OCT-678?atlOrigin=eyJpIjoiZDMzY2ZiNGQ0YTFhNGExNDk1MmUyOTJhNjFkMjhlMDAiLCJwIjoiaiJ9)

---

### Checklist:

- [x] Automated tests updated

---

### Tests:
09a1a2b1bfcfed598e67240d904e1f14


api/package.json
```diff
@@ -51,14 +51,14 @@
...
         "copy-webpack-plugin": "^11.0.0",
-        "crypto-random-string": "^3.3.1",
         "cuid": "^3.0.0",
         "file-type": "^16.5.3",
         "html-to-text": "^8.1.0",
         "jsonwebtoken": "^8.5.1",
         "luxon": "^2.3.0",
         "middy": "^0.36.0",
         "nodemailer": "^6.7.2",
+        "otp-generator": "^4.0.1",
```
florin-holhos, line 61: Not sure if we really need a package for generating the verification code. This new package will get outdated as well. I suggest creating our own helper function since it's relatively easy to generate an alphanumeric string containing unique uppercase letters and digits. Something like this should be enough:

`const generateOTPCharacter = (OTP: string, characterSet: string): string => {
    const newCharacter = characterSet[Math.floor(Math.random() * characterSet.length)];

    return OTP.includes(newCharacter) ? generateOTPCharacter(OTP, characterSet) : newCharacter;
};

const generateOTP = (length: number = 10): string => {
    const allowedCharacters = '0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ';

    if (length > allowedCharacters.length) {
        throw Error(
            'OTP length cannot be greater than the alphanumeric character set used for generating it (' +
                allowedCharacters.length +
                ')'
        );
    }

    let OTP = '';

    while (OTP.length < length) {
        OTP += generateOTPCharacter(OTP, allowedCharacters);
    }

    return OTP;
};`

finlay-jisc: Sounds reasonable to me - we might want to use something like [getRandomValues](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues) instead of Math.random() though.

florin-holhos: 'crypto.getRandomValues()' is a Web API but i see it became available from node 15 as 'webcrypto.getRandomValues()' 

code looks something like this:
`
import { webcrypto } from 'crypto';

const generateOTPCharacter = (OTP: string, characterSet: string): string => {
    const randomNumberArray = webcrypto.getRandomValues(new Uint32Array(1));
    const randomIndex = Math.floor(randomNumberArray[0] * Math.pow(2, -32) * characterSet.length);

    const newCharacter = characterSet[randomIndex];

    return OTP.includes(newCharacter) ? generateOTPCharacter(OTP, characterSet) : newCharacter;
};

const generateOTP = (length = 10): string => {
    const allowedCharacters = '0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ';

    if (length > allowedCharacters.length) {
        throw Error(
            `OTP length cannot be greater than the alphanumeric character set used for generating it (${allowedCharacters.length})`
        );
    }

    let OTP = '';

    while (OTP.length < length) {
        OTP += generateOTPCharacter(OTP, allowedCharacters);
    }

    return OTP;
};
`

florin-holhos: done

florin-holhos (commented):
replaced 'crypto-random-string' with 'generateOTP' helper function

finlay-jisc (approved):

PR #390, opened 2023-05-16
OCT-614 - Queued generation of PDFs - Terraform

Harry-Jisc:
This PR created the Terraform for the sqs queue, sns topics, cloud watch alarm. The only thing missing is the link to the sqs queue and the lambda to trigger generation, this can be added either via terraform or manually when this is confirmed tested and working correctly when merging in the second part of this ticket, the code changes [here](https://github.com/JiscSD/octopus/pull/386)

Currently we only have one slack channel, so will need another created to send emails to for testing on int.

We check in our tfvar file, so you will need to add this line to it before running the terraform:
`slack_channel_email = <email_string>`

---

### Acceptance Criteria:

As Per [OCT-614](https://jiscdev.atlassian.net/browse/OCT-614?atlOrigin=eyJpIjoiMTI3YTNlNzg0OWQ3NGU2ZWE5NmE4NzFhYjA4OGY2MjIiLCJwIjoiaiJ9)

[OCT-614]: https://jiscdev.atlassian.net/browse/OCT-614?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #413, opened 2023-06-12
Set up @octopus.ac inboxes

shaunajisc:
The purpose of this PR was to create the terraform files which will build the infrastructure so we can forward on inbound emails from @octopus.ac addresses to internal emails and slack channels. 

This PR contains the files necessary to build the lambda functions, the SES reciept rule, IAM policy and the s3 action rules. The S3 bucket is in a [separate PR](https://github.com/JiscSD/octopus/pull/407/files) and is called `email_forwarding_bucket`. The infrastructure is set up like this: 

image

---

### Areas of uncertainty 

- I'm not sure if the file structure is correct and if I need to import the email file as a module elsewhere? 
- The documentation and c4 diagrams need to be updated (I think). I will create a separate ticket to update them if so. 

---

### Acceptance Criteria:

See [OCT-631](https://jiscdev.atlassian.net/jira/software/projects/OCT/boards/836?selectedIssue=OCT-631)

[OCT-631]: https://jiscdev.atlassian.net/browse/OCT-631?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


infra/modules/emails/main.tf
```diff
@@ -0,0 +1,102 @@
+variable "email_domains" {
```
Harry-Jisc, line 1: ideally, we don't want these emails to be checked in. if this was closed source it wouldn't be too bad, but since we are open source bots will look through repos and potentially end up spamming these emails. better to have them as variables. 

had to do the same here: https://github.com/JiscSD/octopus/pull/390/commits/23ede00c68f3a770204523e345bc5dfc75764892

terraform will ask you for them at the plan/apply stage or you can directly include them in int/prod.tfvars but don't check them in etc.

infra/create-app/main.tf
```diff
@@ -74,3 +74,8 @@ module "ses" {
     source = "../modules/ses"
     environment = local.environment
 }
+
+module "emails" {
+    source = "../modules/emails"
+    environment = local.environment
```
Harry-Jisc, line 80: this is not used, if you want to use it in the module you'll need to create the variable in a vars.tf file understand module and populate it like so:
```terraform
variable "environment" {
  type = string
}
```

PR #414, opened 2023-06-13
OCT-658 - Affiliations conversion script

finlay-jisc:
The purpose of this PR was to create a script that will convert affiliations stored in the old way (against the publication) to the new way (against the corresponding author).

Note that if this changes a draft/locked publication, if the publication's edit page is viewed, the moved affiliations will be erased. After discussing this behaviour with Florin, it was clear that this will just work for live publications only.

I have tested this by adding some affiliations to publications in the seed data (at least one with coauthors, and one without), adding a bit to the script before the end to console.log the changed publications, then running the script to confirm that the publications have their affiliations moved to the corresponding author.

---

### Acceptance Criteria:

All existing publications that have affiliations listed against the publication as a whole have their affiliations shifted to relate to the corresponding author
If possible, this should also be done for draft/locked publications (if they only have old style affiliations)

Fields mapped to the new affiliations structure are:

- ROR ID (if present)
- “Name” is populated by “Title”
- City
- Country
- Link
- “Affiliation Type” is populated with “Misc.”
- “Disambiguation Source” is populated with “ROR” where a ROR ID is present
---

### Checklist:


api/src/lib/interface.ts
```diff
@@ -670,15 +670,16 @@ export interface OrcidAffiliationDate {
...
     affiliationType:
         | 'membership'
         | 'service'
         | 'invited-position'
         | 'distinction'
         | 'employment'
         | 'education'
-        | 'qualification';
+        | 'qualification'
+        | 'misc.';
```
florin-holhos, line 682: what exactly is this? Do we need it?

finlay-jisc: I added it because of the AC: “Affiliation Type” is populated with “Misc.”
So any new-style affiliation created by this script will have this type.

florin-holhos: hmm, ok. This interface is also used on the frontend to map over and display affiliations. Might worth checking if it won't break the UI

finlay-jisc: It looked OK when I viewed one. The value was displayed and capitalised i.e. "Misc."

api/src/lib/interface.ts
```diff
@@ -670,15 +670,16 @@ export interface OrcidAffiliationDate {
  * - Qualifications
  */
 export interface MappedOrcidAffiliation {
-    id: number;
+    id: number | string;
```
florin-holhos, line 673: I suggest choosing one or the other, but since affiliations coming from ORCID already have a numeric id, and that's what we're going to use from now on, maybe we can generate numeric ids for the old affiliations as well. Keep in mind that this interface is also used on the frontend.

florin-holhos (approved):

florin-holhos (approved):

PR #453, opened 2023-07-25
OCT-713 - Bookmarking a topic

finlay-jisc:
The purpose of this PR was to allow users to bookmark topics.

"As a user, I need to be able to bookmark a research topic so I can stay up-to-date with activities relating to it."

- Standardised publication and topic bookmarks into one generic bookmarks entity
  - This requires a DB migration
- Added script to convert old publication bookmarks to the new style. We can remove the old model later.

---

### Acceptance Criteria:

1. Topics can be bookmarked by clicking the icon in the same manner as other publications
2. Topics show in a list of items displayed under the “My Bookmarks” area, below a “My Bookmarked Topics” heading
    a. The topics list is identical to the bookmarked publications area (except the “topics” in its title)

---

### Checklist:

- [x] Automated tests added

---

### Tests:

UI
Screenshot 2023-07-26 135014

API
Screenshot 2023-07-26 134842

E2E
Screenshot 2023-07-26 133751


ui/src/pages/publications/[id]/index.tsx
```diff
@@ -52,7 +52,7 @@ export const getServerSideProps: Types.GetServerSideProps = async (context) => {
 
     try {
         const response = await api.get(`${Config.endpoints.publications}/${requestedId}/bookmark`, token);
-        bookmark = response.data;
+        bookmark = response.data ? true : false;
```
Comment (florin-holhos) [inline ui/src/pages/publications/[id]/index.tsx:55]: if this is meant to be a boolean, then it should be called **isBookmarked** or something similar

ui/src/pages/publications/[id]/index.tsx
```diff
@@ -84,12 +84,16 @@ type Props = {
...
     const router = useRouter();
     const confirmation = Contexts.useConfirmationModal();
-    const [isBookmarked, setIsBookmarked] = React.useState(props.bookmark ? true : false);
+    const [isBookmarked, setIsBookmarked] = React.useState(props.bookmark);
     const [isPublishing, setPublishing] = React.useState<boolean>(false);
     const [approvalError, setApprovalError] = React.useState('');
     const [serverError, setServerError] = React.useState('');
     const [isEditingAffiliations, setIsEditingAffiliations] = React.useState(false);
 
+    useEffect(() => {
```
Comment (florin-holhos) [inline ui/src/pages/publications/[id]/index.tsx:93]: is this useEffect necessary? since the bookmark is already fetched SSR and it comes via props

ui/src/pages/topics/[id]/index.tsx
```diff
@@ -28,6 +29,13 @@ export const getServerSideProps: Types.GetServerSideProps = async (context) => {
         error = message;
     }
 
+    try {
+        const response = await api.get(`${Config.endpoints.topics}/${id}/bookmark`, token);
+        bookmark = response.data ? true : false;
```
Comment (florin-holhos) [inline ui/src/pages/topics/[id]/index.tsx:34]: same as the previous comment, this should follow naming conventions for booleans

florin-holhos (commented):
This approach also works, but if we have 10 types of bookmarks we don't need to create tables, endpoints and services for each type. We can have a single Bookmark table containing all the bookmarks referencing different resources into DB.

For example, all bookmarks could have the following fields:
- id String (unique bookmark id)
- type  Enum "publication" | "topic" | "and_so_on"
- resourceId String - this would be a specific publication/topic id
- userId String
- createdAt DateTime

With this approach, the code will be cleaner and easier to extend if we need to add more logic for bookmarks or new bookmark types

api/prisma/schema.prisma
```diff
@@ -207,19 +207,18 @@ model Topic {
...
-    topicId       String
+    type          BookmarkType
+    entityId      String
     userId        String
     createdAt     DateTime    @default(now())
-    topic         Topic       @relation(fields: [topicId], references: [id], onDelete: Cascade)
     user          User        @relation(fields: [userId], references: [id], onDelete: Cascade)
 
-    @@unique([topicId, userId])
+    @@unique([entityId, type, userId])
```
florin-holhos, line 221: Not sure we need the type included into the unique constraint. As long as we have the userId and the entityId it should be enough

api/serverless.yml
```diff
@@ -415,24 +415,24 @@ functions:
...
-                  path: ${self:custom.versions.v1}/publications/{id}/bookmark
+                  path: ${self:custom.versions.v1}/bookmarks/{id}
                   method: DELETE
                   cors: true
     getBookmark:
         handler: src/components/bookmark/routes.get
         events:
             - http:
-                  path: ${self:custom.versions.v1}/publications/{id}/bookmark
+                  path: ${self:custom.versions.v1}/bookmarks/{type}/{entityId}
```
florin-holhos, line 432: this endpoint should look the same as the DELETE endpoint: /bookmarks/{id}

this endpoint requires the user to be logged in in order to be able to delete a bookmark, that means we have all the info we need: the bookmark id and the user id.

api/src/components/bookmark/controller.ts
```diff
@@ -1,114 +1,214 @@
...
 
-        return response.json(200, { message: 'Bookmark deleted from this publication.' });
+        return response.json(200, { message: 'Bookmark deleted successfully.' });
     } catch (err) {
         return response.json(500, { message: 'Unknown server error.' });
     }
 };
 
+// For checking whether an entity is bookmarked.
 export const get = async (
```
florin-holhos, line 162: this controller should simply return a bookmark by a given id. We can use getAll with some filters in order to check if a specific bookmark exists

api/src/components/bookmark/service.ts
```diff
@@ -1,92 +1,151 @@
...
+            id
         }
     });
 
     return bookmark;
 };
 
-export const getAll = async (userId: string) => {
-    const bookmark = await client.prisma.publicationBookmarks.findMany({
+export const getAll = async (userId: string, type: I.BookmarkType): Promise<I.PopulatedBookmark[]> => {
```
florin-holhos, line 66: we can use this service to filter bookmarks based on our needs, that means it could take an argument of type BookmarkWhereInput and then, from the 'getAll' controller we can provide {userId, type, entityId} as filters. This way, we can check if a specific bookmark exists, by fetching it like this: /bookmarks?type=PUBLICATION&entityId=pub_id_123. This will return an array with a single record if the bookmark exists or an empty array otherwise.

ui/src/pages/my-bookmarks.tsx
```diff
@@ -11,38 +11,62 @@ import * as api from '@api';
...
-    const [userBookmarks, setUserBookmarks] = React.useState(props.usersBookmarks);
+    const [userPublicationBookmarks, setUserPublicationBookmarks] = React.useState(props.userPublicationBookmarks);
+    const [userTopicBookmarks, setUserTopicBookmarks] = React.useState(props.userTopicBookmarks);
 
-    const deleteBookmark = async (publication: string) => {
+    const deletePublicationBookmark = async (id: string) => {
+        try {
+            await api.destroy(`${Config.endpoints.bookmarks}/${id}`, props.token);
+
+            const getPublicationBookmarks = await api.get(
```
florin-holhos, line 55: if the bookmark is deleted successfully, we can simply update the state without it, fetching the bookmarks again is not needed.

ui/src/pages/my-bookmarks.tsx
```diff
@@ -11,38 +11,62 @@ import * as api from '@api';
...
+        }
+    };
+    const deleteTopicBookmark = async (id: string) => {
         try {
-            await api.destroy(`${Config.endpoints.publications}/${publication}/bookmark`, props.token);
+            await api.destroy(`${Config.endpoints.bookmarks}/${id}`, props.token);
 
-            const getBookmarks = await api.get(`${Config.endpoints.bookmarks}`, props.token);
-            setUserBookmarks(getBookmarks.data);
+            const getTopicBookmarks = await api.get(`${Config.endpoints.bookmarks}?type=TOPIC`, props.token);
```
florin-holhos, line 68: same as above, if the bookmark was deleted successfully, we don't need to fetch the bookmark list, we can just update the state.

ui/src/pages/publications/[id]/index.tsx
Comment (finlay-jisc) [reply ui/src/pages/publications/[id]/index.tsx]: Renamed to isBookmarked

ui/src/pages/publications/[id]/index.tsx
Comment (finlay-jisc) [reply ui/src/pages/publications/[id]/index.tsx:95]: It's needed to prevent the bookmark icon incorrectly keeping its state when navigating from one publication / topic to another

api/serverless.yml
finlay-jisc: I removed this endpoint since the UI won't be using it. As you recommended below, it now hits getMany (previously getAll) with filters to check if an entity is bookmarked

api/src/components/bookmark/controller.ts
```diff
@@ -1,114 +1,214 @@
...
-                message: 'You have already deleted the bookmark for this publication.'
+        // Check that the user owns the bookmark
+        if (bookmark.userId !== event.user.id) {
+            return response.json(403, {
+                message: 'You do not own this bookmark.'
             });
         }
 
-        await bookmarkService.remove(bookmark.id);
+        // Checks depending on bookmark type
```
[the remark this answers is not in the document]
finlay-jisc: That's true, removed now

api/src/components/bookmark/controller.ts
finlay-jisc: Removed this function as the UI doesn't need it anymore.

api/src/components/bookmark/service.ts
finlay-jisc: Makes sense, I've changed it to allow an entityId filter too

ui/src/pages/my-bookmarks.tsx
finlay-jisc: Yes, corrected now

api/src/components/bookmark/controller.ts
```diff
@@ -1,134 +1,130 @@
...
-    event: I.AuthenticatedAPIRequest<undefined, undefined, I.GetAllBookmarkPathParams>
+    event: I.AuthenticatedAPIRequest<undefined, I.GetAllBookmarksQueryStringParameters>
 ): Promise<I.JSONResponse> => {
     try {
-        const bookmarks = await bookmarkService.getAll(event.user.id);
+        const typeFilter: I.BookmarkType = event.queryStringParameters.type;
+        const entityFilter = event.queryStringParameters.entityId || undefined;
+        const bookmarks = await bookmarkService.getMany(event.user.id, typeFilter, entityFilter);
 
         if (!bookmarks) {
```
florin-holhos, line 125: This if statement will never execute since 'bookmarkService.getMany' will always return an array, even if it returns an empty array, !bookmarks will always be false. I think we can safely remove this check and simply return the empty array to the client if that's the case.

ui/src/pages/publications/[id]/index.tsx
```diff
@@ -77,19 +77,25 @@ export const getServerSideProps: Types.GetServerSideProps = async (context) => {
...
     userToken: string;
 };
 
 const Publication: Types.NextPage<Props> = (props): React.ReactElement => {
     const router = useRouter();
     const confirmation = Contexts.useConfirmationModal();
-    const [isBookmarked, setIsBookmarked] = React.useState(props.bookmark ? true : false);
+    const isBookmarked = props.bookmarkId ? true : false;
+    const [showBookmark, setShowBookmark] = React.useState(isBookmarked);
+    const [bookmarkId, setBookmarkId] = React.useState(props.bookmarkId);
```
Comment (florin-holhos) [inline ui/src/pages/publications/[id]/index.tsx:89]:
```suggestion
    const [bookmarkId, setBookmarkId] = React.useState(props.bookmarkId);
    const isBookmarked = bookmarkId ? true : false;
```
There's a small bug with the bookmarking mechanism while bookmarking a publication and navigating to a new one, the new publication shows as bookmarked as well until the page is refreshed. 
There's another small bug when trying to remove a bookmark, sometimes the bookmarkId is null because the state was not updated.

You may not need 'showBookmark' state since you already have 'bookmarkId' state. We just need to make sure that 'bookmarkId' is updated properly.

api/src/components/bookmark/controller.ts
finlay-jisc: Well spotted, thanks

ui/src/pages/publications/[id]/index.tsx
```diff
@@ -200,19 +206,32 @@ const Publication: Types.NextPage<Props> = (props): React.ReactElement => {
     );
 
     const onBookmarkHandler = async () => {
-        if (isBookmarked) {
+        if (showBookmark) {
```
Comment (florin-holhos) [inline ui/src/pages/publications/[id]/index.tsx:209]:
```suggestion
        if (isBookmarked) {
```

We can replace with 'isBookmarked'

ui/src/pages/publications/[id]/index.tsx
```diff
@@ -200,19 +206,32 @@ const Publication: Types.NextPage<Props> = (props): React.ReactElement => {
...
             //delete the bookmark
             try {
-                await api.destroy(`publications/${publicationData?.id}/bookmark`, user?.token);
-                setIsBookmarked(false);
+                await api.destroy(`bookmarks/${bookmarkId}`, user?.token);
+                setShowBookmark(false);
```
Comment (florin-holhos) [inline ui/src/pages/publications/[id]/index.tsx:213]:
```suggestion
                setBookmarkId(null);
```

We can replace with 'setBookmarkId'

ui/src/pages/publications/[id]/index.tsx
```diff
@@ -200,19 +206,32 @@ const Publication: Types.NextPage<Props> = (props): React.ReactElement => {
...
+                    userId: string;
+                }>(
+                    `bookmarks`,
+                    {
+                        type: 'PUBLICATION',
+                        entityId: publicationData?.id
+                    },
+                    user?.token
+                );
+                setShowBookmark(true);
```
Comment (florin-holhos) [inline ui/src/pages/publications/[id]/index.tsx:233]:
```suggestion
```

This is not needed anymore

ui/src/pages/publications/[id]/index.tsx
```diff
@@ -502,17 +521,17 @@ const Publication: Types.NextPage<Props> = (props): React.ReactElement => {
...
                             </h1>
-                            {isBookmarkVisible && (
+                            {isBookmarkButtonVisible && (
                                 <div className="col-span-1 grid justify-items-end">
                                     <button
                                         className="h-8 hover:cursor-pointer focus:outline-none focus:ring focus:ring-yellow-200 focus:ring-offset-2 dark:outline-none dark:focus:ring dark:focus:ring-yellow-600 dark:focus:ring-offset-1"
                                         onClick={onBookmarkHandler}
                                         aria-label="toggle-bookmark"
-                                        title={`${isBookmarked ? 'Remove bookmark' : 'Bookmark this publication'}`}
+                                        title={`${showBookmark ? 'Remove bookmark' : 'Bookmark this publication'}`}
```
Comment (florin-holhos) [inline ui/src/pages/publications/[id]/index.tsx:530]:
```suggestion
                                        title={`${isBookmarked ? 'Remove bookmark' : 'Bookmark this publication'}`}
```

we can replace with 'isBookmarked'

ui/src/pages/publications/[id]/index.tsx
```diff
@@ -502,17 +521,17 @@ const Publication: Types.NextPage<Props> = (props): React.ReactElement => {
...
                                     >
                                         <OutlineIcons.BookmarkIcon
                                             className={`h-8 w-8 ${
-                                                isBookmarked ? 'fill-blue-700 dark:fill-blue-50' : 'fill-transparent'
+                                                showBookmark ? 'fill-blue-700 dark:fill-blue-50' : 'fill-transparent'
```
Comment (florin-holhos) [inline ui/src/pages/publications/[id]/index.tsx:534]:
```suggestion
                                                isBookmarked ? 'fill-blue-700 dark:fill-blue-50' : 'fill-transparent'
```

we can replace with 'isBookmarked'

ui/src/pages/topics/[id]/index.tsx
```diff
@@ -36,21 +44,73 @@ export const getServerSideProps: Types.GetServerSideProps = async (context) => {
...
+    const isBookmarked = props.bookmarkId ? true : false;
+    const [showBookmark, setShowBookmark] = React.useState(isBookmarked);
+    const [bookmarkId, setBookmarkId] = React.useState(props.bookmarkId);
+
     const topic = props.topic;
     const showChildren = Boolean(topic.children.length);
     const showParents = Boolean(topic.parents.length);
+
+    useEffect(() => {
+        setShowBookmark(isBookmarked);
```
Comment (florin-holhos) [inline ui/src/pages/topics/[id]/index.tsx:70]: On this page we have the same bug regarding bookmarking mechanism. The same fix what i suggested for publications can be applied here also.

florin-holhos (approved):

florin-holhos (approved):

PR #467, opened 2023-08-17
OCT-620: Adding links to topics when creating a research problem

finlay-jisc:
The purpose of this PR was: As a corresponding author, I need to have clarity on where to attach my research problem if I am not deriving it from an existing Octopus publication; in this case, I should be prompted to add my publication to a topic.

To note:

- In order for the HTML to make sense semantically I've positioned the dropdown above the combobox, instead of to the left of it
- Some names have changed (e.g. the tab from "Linked Publications" to "Linked items") - this has been discussed with Tim
- I cherry-picked a commit from the topics search branch because it made sense to use the getTopics API endpoint from that

---

### Acceptance Criteria:

As per [OCT-620](https://jiscdev.atlassian.net/browse/OCT-620?atlOrigin=eyJpIjoiMWRiODBkOGMyMTExNDc3Y2I1NTQwOGM5Njg0YzE1NWUiLCJwIjoiaiJ9)

---

### Checklist:

- [x] Automated tests added
- [x] Documentation updated (guidance text on page)

---

### Tests:
E2E:
Screenshot 2023-08-17 155816

API:
Screenshot 2023-08-17 160354

UI:
Screenshot 2023-08-17 160457

[OCT-620]: https://jiscdev.atlassian.net/browse/OCT-620?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/src/lib/interface.ts
```diff
@@ -138,6 +138,15 @@ export interface UpdatePublicationRequestBody {
     dataPermissionsStatementProvidedBy?: string;
     dataAccessStatement?: string;
     selfDeclaration?: boolean;
+    topics?: string[];
+}
+
+export interface UpdatePublicationData extends Omit<UpdatePublicationRequestBody, 'topics'> {
```
florin-holhos, line 144: I think it's a bit odd to first define the "topics" field on UpdatePublicationRequestBody and then extending that interface but omitting "topics" field which you just defined, just to change it's type to something that prisma expects. I would rather use UpdatePublicationRequestBody interface as it is, with the "topics" as a string[] and do that set mapping into the update service itself.

api/src/components/publication/controller.ts
```diff
@@ -224,7 +220,13 @@ export const update = async (
             });
         }
 
-        const updatedPublication = await publicationService.update(event.pathParameters.id, event.body);
+        const updateContent = {
+            ...event.body,
+            ...(event.body.content && { content: helpers.getSafeHTML(event.body.content) }),
+            ...(event.body.topics && { topics: { set: event.body.topics.map((topicId) => ({ id: topicId })) } })
+        } as I.UpdatePublicationData;
```
florin-holhos, line 227: if you move the topics ids mapping inside the update service, then you don't need to cast as I.UpdatePublicationData anymore.

ui/src/components/Publication/Creation/LinkedItems/index.tsx
```diff
@@ -0,0 +1,169 @@
...
+                </p>
+
+                <p className="mb-6 block text-sm text-grey-800 transition-colors duration-500 dark:text-white-50">
+                    This approach structures the content of the platform and makes your work more discoverable. It also
+                    helps you avoid repeating work, as there is no need to keep re-writing descriptions of well-known
+                    Research Problems, or Methods etc.
+                </p>
+            </div>
+
+            <div className="relative">
```
florin-holhos, line 113: This layout needs some additional css based on ticket description no. 4: _"To the right of the dropdown, a search is present. The selected item type in the dropdown determines what item type will display in the search box’s results preview"_

The UI should look something like this from what I understand:
image

But it looks like this:
image

ui/src/lib/interfaces.ts
```diff
@@ -500,3 +500,18 @@ export interface TopicsPaginatedResults {
 export interface TopicsPageQuery extends ParsedUrlQuery {
     query?: string;
 }
+
+export interface TopicsPaginatedResults {
```
florin-holhos, line 504: This is a duplicate. It was already defined on line 489. It might come from rebase.

finlay-jisc: Now removed.

api/src/lib/interface.ts
finlay-jisc: That makes sense - I've removed this new interface and moved the topics manipulation to the service file.

api/src/components/publication/controller.ts
finlay-jisc: Agreed, moved the mapping to the service and removed this cast.

ui/src/components/Publication/Creation/LinkedItems/index.tsx
finlay-jisc: Fixed the position now with your help!

florin-holhos (dismissed):

florin-holhos (approved):

PR #471, opened 2023-08-30
OCT-620 - stop non-problem publications being linkable to topics

finlay-jisc:
The purpose of this PR was to correct a bug with the original work on OCT-620. Only research problems should have the option to link to a topic, but in testing, any publication type could do this.

---

### Acceptance Criteria:

Only research problems can be linked to a topic using the UI.

---

### Checklist:

- [x] Automated tests added
- [ ] Documentation updated

### DIFF
--- a/ui/src/components/Publication/Creation/LinkedItems/index.tsx
+++ b/ui/src/components/Publication/Creation/LinkedItems/index.tsx
@@ -113,7 +113,7 @@ const Links: React.FC = (): React.ReactElement => {
             <div className="relative">
                 <Components.PublicationCreationStepTitle text="Add links" required />
                 <div className="flex flex-col flex-wrap gap-4 sm:flex-row sm:items-center">
-                    {!!isProblem && (
+                    {isProblem && (
                         <select
                             name="linked-entity-type"
                             id="linked-entity-type"
@@ -129,7 +129,7 @@ const Links: React.FC = (): React.ReactElement => {
                         </select>
                     )}
                     <div className="flex-1">
-                        {entityType === 'TOPIC' && !!isProblem ? (
+                        {entityType === 'TOPIC' ? (
                             <Components.LinkedTopicsCombobox
                                 fetchAndSetLinks={fetchAndSetLinks}
                                 setError={setError}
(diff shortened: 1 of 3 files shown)


ui/src/components/Publication/Creation/LinkedItems/index.tsx
```diff
@@ -113,34 +113,36 @@ const Links: React.FC = (): React.ReactElement => {
...
-                            const value: Types.LinkedEntityType = e.target.value as Types.LinkedEntityType;
-                            setEntityType(value);
-                        }}
-                        value={entityType}
-                        className="block rounded-md border border-grey-200 outline-none focus:ring-2 focus:ring-yellow-500 sm:mr-0"
-                    >
-                        <option value="PUBLICATION">Publications</option>
-                        <option value="TOPIC">Research topics</option>
-                    </select>
+                    {!!isProblem && (
```
florin-holhos, line 116: "isProblem" is already a boolean. Does it need to be converted to a boolean again?

ui/src/components/Publication/Creation/LinkedItems/index.tsx
```diff
@@ -156,7 +158,7 @@ const Links: React.FC = (): React.ReactElement => {
                     entityType="PUBLICATION"
                 />
             )}
-            {!error && !!topics.length && (
+            {!error && !!topics.length && !!isProblem && (
```
florin-holhos, line 161: Topics length will always be 0 for other publication types other than PROBLEM.
'&& !!isProblem' shouldn't make any difference.

florin-holhos (changes requested):
On the API side, we need to make sure we don't allow people to add topics to other publication types beside PROBLEM. Whenever we create/update a publication, if we get topics into the request body, we need to check if the publication types is PROBLEM, otherwise the API should return an error. I managed to add topics to a Hypothesis using Postman.

ui/src/components/Publication/Creation/LinkedItems/index.tsx
finlay-jisc: No, there is no need

PR #480, opened 2023-09-11
OCT-752: Ensure API tests pass

finlay-jisc:
The purpose of this PR was to get the versioning code in a state where the API tests pass, without changing the tests unless really necessary.

This includes another migration but it doesn't mean we have to change more things again. It just corrects some relations that should have been mandatory.

I tried using the app a little bit and most things seem to work, but I noticed that saved links cause an error when you try to view them. I'm sure there will be other bugs but hopefully the end to end tests will catch them.

---

### Acceptance Criteria:

As per [OCT-752](https://jiscdev.atlassian.net/browse/OCT-752).

---

### Checklist:

- [ ] Automated tests added - a couple of minimal changes to API tests
- [ ] Documentation updated

---

### Tests:

Screenshot 2023-09-11 151524

[OCT-752]: https://jiscdev.atlassian.net/browse/OCT-752?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/prisma/migrateToVersionedPublications.ts
```diff
@@ -1,3 +1,5 @@
+// Disable TS checking as this is only used against an outdated prisma schema
+// @ts-nocheck
```
florin-holhos, line 2: Disabling TS checking on this file can lead to data loss while migrating data from one model to another, especially for optional fields and relations. I think it would have been better/safer to rename the old Publication model first, then create the new models so that we have TS type checking, autocomplete etc for the old model and for the new models as well. Then we can create new publications and their v1 while looping through the old ones and mapping all the necessary fields/relations. After that we can simply drop the old (renamed) Publication model and it's relations.

api/prisma/reindex.ts
```diff
@@ -15,30 +15,48 @@ const reindex = async () => {
...
-                licence: pub.licence,
-                description: pub.description,
-                keywords: pub.keywords,
-                content: pub.content,
-                language: 'en',
-                currentStatus: pub.currentStatus,
-                publishedDate: pub.publishedDate,
-                cleanContent: htmlToText.convert(pub.content)
-            }
+        const latestLiveVersion = pub.versions.find((version) => {
```
florin-holhos, line 36: We can do this filtering into the prisma query directly to only include the last LIVE version of each publication. While running this script the search functionality is down. We want to keep it running as fast as possible.

api/prisma/migrateToVersionedPublications.ts
finlay-jisc: What I could do is change the migration instructions to instruct that only the specific commit is checked out, then TS checking will be enabled at the point the script is run (this was the case when developing and testing the migration script, so we know the TS checking works). After the "versioning_post_migration" prisma migration is run, we wouldn't have any need for this script. It may actually be better to remove the script entirely instead of marking it as ts-nocheck. I'll do that instead and edit the migration instructions

api/src/components/publication/controller.ts
```diff
@@ -229,7 +221,9 @@ export const update = async (
             });
         }
 
-        const updatedPublication = await publicationService.update(event.pathParameters.id, event.body);
+        await publicationService.update(event.pathParameters.id, event.body);
```
florin-holhos, line 224: This is the part where it starts to get confusing, because we call a Publication service that is supposed to update a Publication, but we update a PublicationVersion instead, which is a different entity into our DB. All the checks above, are done against the publication, but the publication itself doesn't have any of those fields beside the "type". I wonder how we can improve these controllers and make it more clear what we are checking and what we are updating. I think the "update" controller and service should be renamed to "updateLatestVersion" so we know exactly what they are doing and all those checks above we can split them as follows:

const publication = await publicationService.get(event.pathParameters.id);
const latestVersion = publication.versions[0]; // get service includes latest version by default
// do all the necessary checks using "publication" and "latestVersion" accordingly

api/src/components/reference/controller.ts
```diff
@@ -40,7 +40,7 @@ export const updateAll = async (
         }
 
         // check that we are updating the publication version with the correct id
-        if (event.body.some((reference) => reference.publicationVersionId !== publication.versionId)) {
+        if (event.body.some((reference) => reference.publicationId !== publication.id)) {
```
florin-holhos, line 43: Is this supposed to be "reference.publicationVersionId" ? since the references are stored on the PublicationVersion entity

florin-holhos: I guess the UI will need some changes as well... 
The check should look like this:
reference.publicationVersionId !== publication.versions[0].id

api/src/components/reference/service.ts
```diff
@@ -17,18 +18,43 @@ export const getAllByPublication = async (publicationId: string) => {
         }
     });
 
-    return references;
+    // Return with publicationId.
+    return references.map((reference) => {
+        const { publicationVersionId, ...referenceRest } = reference;
+
+        return { publicationId, ...referenceRest };
```
florin-holhos, line 25: Is the "publicationId" really needed? The return type of this service doesn't match any model into our DB if we combine it like that. We will end up with scenarios where we need to get rid of this "publicationId"

api/src/components/reference/service.ts
```diff
@@ -17,18 +18,43 @@ export const getAllByPublication = async (publicationId: string) => {
...
+
     await client.prisma.references.deleteMany({
         where: {
-            publicationVersionId
+            publicationVersionId: latestVersion?.id
         }
     });
 
+    // Event body contains a publication ID. Replace this with the latest version's ID.
+    const createManyInput = data.map((reference) => {
```
florin-holhos, line 51: Instead of doing all these spreads while looping, just to keep the UI happy, i think we should simply change the "publicationId" field with "publicationVersionId" on the UI side. This will also reflect what we have into DB

api/src/scripts/updateDoi.ts
```diff
@@ -209,7 +209,12 @@ const updateDoi = async (): Promise<void> => {
     let index = 1;
 
     for (const publication of simplifiedPublications) {
-        await helpers.updateDOI(publication.doi, publication, publication.References).catch((err) => console.log(err));
+        const abstractedReferences = publication.References.map((reference) => {
+            const { publicationVersionId, ...referenceRest } = reference;
+
+            return { publicationId: publication.id, ...referenceRest };
```
florin-holhos, line 215: The "publicationId" is not used by the _updateDOI_ helper. There's no need to map the references here

api/prisma/reindex.ts
finlay-jisc: I can't think how to do it in one query - if there was a separate query for each publication you could do something like
```
client.prisma.publicationVersions.findFirst({
    where: {
        versionOf: publicationId,
        currentStatus: 'LIVE',
    },
    orderBy: {
        versionNumber: 'desc'
    }
});
```
But because we can't limit the "include" to only return one result like we can with a findFirst, I don't know how we can do it in the single query above.

finlay-jisc: Actually, we can order the "included" versions by versionNumber descending and then access the latest one at versions[0]. I will do that

api/src/components/publication/controller.ts
finlay-jisc: I've pushed a commit just now that hopefully improves this, I agree that it's a clearer way of showing what exactly we're dealing with. I've not fully updated references because I am going to address those next following the other comments and make them deal directly with versions (like coauthors do)

api/prisma/reindex.ts
```diff
@@ -15,31 +15,51 @@ const reindex = async () => {
...
+            }
+        },
+        include: {
+            versions: {
+                where: {
+                    currentStatus: 'LIVE'
+                },
+                orderBy: {
+                    versionNumber: 'desc'
+                }
```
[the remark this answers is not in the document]
finlay-jisc: That would exclude publications where there is a live version but also an active draft. It could look like this for example:
```
...
versions: {
    ...
    {
        ...
        versionNumber: 2,
        isCurrent: false,
        currentStatus: 'LIVE'
    },
    {
        ...
        versionNumber: 3,
        isCurrent: true,
        currentStatus: 'DRAFT'
    }
}
```
This wouldn't be returned if we used that where clause.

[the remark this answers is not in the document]
florin-holhos: That where clause I suggested only applies to versions, not to the publications. It filters the versions and only returns the ones where "currentStatus=LIVE" and "isCurrent=true" which means it will only return one version, instead of all of them in a reverse order. It should run faster.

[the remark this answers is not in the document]
florin-holhos: In the example you have provided the data is wrong. If we have v2 and v3, and v3 is a DRAFT, then v2 MUST be the latest LIVE version. Otherwise our system is broken, because it should only allow creating a new version from the latest LIVE version which will continue to be the latest until the new DRAFT becomes LIVE.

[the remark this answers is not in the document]
finlay-jisc: I get it - I would do it that way as well, but it would still exclude versions (I should have said versions above, not publications) like version 2 in the example I put above. Because there is a new draft (which would have isCurrent = true, and only one version can have this), the latest live version no longer has isCurrent = true. But the latest live version is the one we want to put into the index. So we can't only include versions that have isCurrent = true.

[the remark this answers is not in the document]
florin-holhos: Closing this due to lack of info on how we are going to implement the new version creation mechanism. We may revisit it later on after we have an implementation of OC-391

api/src/components/flag/controller.ts
```diff
@@ -53,13 +53,23 @@ export const createFlag = async (
     try {
         const publication = await publicationService.get(event.pathParameters.id);
 
-        if (!publication || publication.currentStatus !== 'LIVE') {
+        if (!publication || !publication.versions.some((version) => version.currentStatus === 'LIVE')) {
             return response.json(404, {
-                message: 'Cannot flag that a publication that does not exist, or is not LIVE'
+                message: 'Cannot flag that a publication that does not exist, or has never been LIVE'
```
florin-holhos, line 58: May need a rewording for `Cannot flag that a publication that...`

api/src/components/flag/controller.ts
```diff
@@ -53,13 +53,23 @@ export const createFlag = async (
...
             });
         }
 
-        if (publication.user.id === event.user.id) {
+        // Latest published version (the version presented when someone tries to raise a flag) is either current version,
+        // or if that's not LIVE, the version before, as we've already confirmed that the publication has been live at some point.
+        const latestPublishedVersion =
+            publication.versions.find((version) => version.isCurrent && version.currentStatus === 'LIVE') ||
+            publication.versions.find((version) => version.versionNumber === publication.versions.length - 1);
```
florin-holhos, line 66: Is the second condition (after || ) really needed? It allows users to flag DRAFT versions.

api/src/components/flag/controller.ts
```diff
@@ -53,13 +53,23 @@ export const createFlag = async (
     try {
         const publication = await publicationService.get(event.pathParameters.id);
 
-        if (!publication || publication.currentStatus !== 'LIVE') {
+        if (!publication || !publication.versions.some((version) => version.currentStatus === 'LIVE')) {
```
florin-holhos, line 56: Since flags can only be raised against the latest published version, this condition needs to check whether the latest LIVE version of this publication exists.

api/src/components/flag/controller.ts
```diff
@@ -153,8 +163,18 @@ export const createFlagComment = async (
...
         }
 
-        // The user attempting to leave a comment, is not the flag creator, or the publication owner
-        if (flag.createdBy !== event.user.id && flag.publication.user.id !== event.user.id) {
+        const publication = flag.publication;
+
+        // The user attempting to leave a comment is not the flag creator or the publication owner.
+        const latestPublishedVersion =
+            publication.versions.find((version) => version.isCurrent && version.currentStatus === 'LIVE') ||
+            publication.versions.find((version) => version.versionNumber === publication.versions.length - 1);
```
florin-holhos, line 171: the same as above, we can only allow users to flag the latest published version

api/src/components/flag/controller.ts
```diff
@@ -236,17 +254,25 @@ export const resolveFlag = async (
 
         const resolveFlag = await flagService.resolveFlag(event.pathParameters.id);
 
-        const publication = await publicationService.get(flag.publicationId);
+        const publication = flag.publication;
+
+        const latestPublishedVersion =
+            publication.versions.find((version) => version.isCurrent && version.currentStatus === 'LIVE') ||
+            publication.versions.find((version) => version.versionNumber === publication.versions.length - 1);
```
florin-holhos, line 261: Flags against DRAFT versions are not allowed, we only need the first condition before ||

api/src/components/reference/controller.ts
finlay-jisc: This and the other issues with references using publication ID etc. should be fixed since commit [b1578e7](https://github.com/JiscSD/octopus/pull/480/commits/b1578e743e7789735090a7f54d6d049d37f73d7f)

api/src/components/publication/controller.ts
```diff
@@ -103,7 +104,13 @@ export const deletePublication = async (
             });
         }
 
-        if (publication.user.id !== event.user.id) {
+        const firstVersion = publication.versions.find((version) => version.versionNumber === 1);
```
florin-holhos, line 107: After discussing with PO about this flow, only the last version can be deleted (by it's creator) while it's still a DRAFT. If there's only one version, we delete the publication itself.

api/src/components/bookmark/controller.ts
```diff
@@ -13,26 +13,37 @@ export const create = async (
...
+                if (!publication.versions.some((version) => version.currentStatus === 'LIVE')) {
                     return response.json(403, {
-                        message: 'You cannot bookmark a draft publication.'
+                        message: 'You cannot bookmark a publication which has not gone live.'
                     });
                 }
 
-                // Check to see if the user is the author or co author. If so throw an error
-                const isUserCoAuthor = publication.coAuthors.some((coAuthor) => coAuthor.linkedUser == event.user.id);
+                const latestPublishedVersion =
```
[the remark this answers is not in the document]
finlay-jisc: Have removed this check (and associated tests) now.

api/src/scripts/updateDoi.ts
```diff
@@ -198,19 +198,12 @@ const updateDoi = async (): Promise<void> => {
...
-        return { ...versionRest, versionId: versionRest.id, ...publication };
-    });
-
     let index = 1;
 
-    for (const publication of simplifiedPublications) {
-        await helpers.updateDOI(publication.doi, publication, publication.References).catch((err) => console.log(err));
-        console.log(`No: ${index}. ${publication.title} doi updated (${publication.doi})`);
+    for (const publication of publications) {
+        const references = publication.versions[0].References;
```
florin-holhos, line 204: publications.versions[0] may not be the latest published version

api/src/components/flag/controller.ts
finlay-jisc: This has been addressed with the new isLatestLiveVersion field now.

api/src/scripts/updateDoi.ts
finlay-jisc: This should be OK now that we're using isLatestLiveVersion in this file

api/src/components/publication/controller.ts
finlay-jisc: Pushed a commit that made this a bit clearer - we can wait to write the code to delete a version later when we have a way to add more than one version.

florin-holhos (approved):

PR #485, opened 2023-09-15
OC-95: render meta tags at document level instead of client side js

finlay-jisc:
The purpose of this PR was to fix an issue where although meta tag values were correct when inspecting with devtools, twitter cards and opengraph testing utilities weren't picking them up. I realised this is because in the "view source" (the document without javascript being run) of each page, the meta tag values were still the defaults set at the app level. The way next JS was overriding them involved client side JS. I've changed the method of setting the meta tag values (for og:title and og:description) so that they will be set correctly in the "view source" version of a page.

---

### Acceptance Criteria:

As per [OC-95](https://jiscdev.atlassian.net/browse/OC-95)

---

### Checklist:

- [ ] Automated tests added
- [ ] Documentation updated

---

### Tests:

UI:
Screenshot 2023-09-15 135123

E2E:
Screenshot 2023-09-15 134712

[OC-95]: https://jiscdev.atlassian.net/browse/OC-95?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ui/src/pages/_app.tsx
```diff
@@ -15,13 +15,24 @@ import '../styles/globals.css';
...
+        description?: string;
+    };
 };
 
 const App = ({ Component, pageProps }: Types.AppProps<CustomProps>) => {
     const [mounted, setMounted] = useState(false);
     const { user } = Stores.useAuthStore();
     const { darkMode } = Stores.usePreferencesStore();
 
+    const metadata = {
```
florin-holhos, line 29: The code inside _app.tsx will execute on every route/subroute change, even when adding query params into the URL. I don't recommend adding code into this file unless it's absolutely necessary. I recommend leaving this responsibility (of rendering proper meta tags) at the page level as it was. Since you already added that "metadata" prop which is computed server side, the tags can be rendered with the info coming via props on each page.

florin-holhos (dismissed):
Approved as a temporary solution. A lot of this work will be revisited in [OC-700](https://jiscdev.atlassian.net/browse/OC-700) where we will enable SSR for pages and the meta tags will work properly

florin-holhos (approved):

PR #488, opened 2023-09-20
OC-396: Ensure end to end tests pass

finlay-jisc:
The purpose of this PR was to check that the end to end tests still work properly after versioning has been introduced.

Mostly it was adjusting the UI to accommodate the API changes made so far, but I also changed the coauthor API routes and the my-affiliations API route to take a version ID instead of a publication ID (as we've done for References).

I did notice an issue where, when viewing a published hypothesis, the visualisation was showing an unexpected problem (as well as the one problem the hypothesis was linked to). Since this seems to be an issue with the API response from the get links function which is being worked on in parallel, I think we should check if it's still a bug after that work is done, since it might fix it.

---

### Acceptance Criteria:

The end to end test suite passes on the versioning branch.

---

### Checklist:

- [ ] Automated tests added - none added but some changes
- [ ] Documentation updated

---

### Tests:

UI:
Screenshot 2023-09-20 131611

API:
Screenshot 2023-09-20 132032

E2E:
Screenshot 2023-09-20 130011

### DIFF
--- a/api/src/components/affiliations/controller.ts
+++ b/api/src/components/affiliations/controller.ts
@@ -15,17 +15,17 @@ export const updateAffiliations = async (
         const isIndependent = event.body.isIndependent;
         const affiliations = event.body.affiliations;
         const publicationVersionId = event.pathParameters.id;
-        // Get publication with current version
+        // Get publication version
         const version = await publicationVersionService.get(publicationVersionId);
 
-        //check that the publication exists
+        // Check that the version exists
         if (!version) {
             return response.json(404, {
                 message: 'This publication does not exist.'
             });
         }
 
-        //check if publication is LIVE
+        // Check if version is LIVE
         if (version.currentStatus === 'LIVE') {
             return response.json(403, {
                 message: 'You cannot add affiliations to a LIVE publication.'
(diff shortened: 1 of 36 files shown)


api/src/components/affiliations/controller.ts
```diff
@@ -14,27 +14,25 @@ export const updateAffiliations = async (
     try {
         const isIndependent = event.body.isIndependent;
         const affiliations = event.body.affiliations;
-        const publicationId = event.pathParameters.id;
+        const publicationVersionId = event.pathParameters.id;
         // Get publication with current version
-        const publication = await publicationService.getWithVersion(publicationId);
+        const version = await publicationVersionService.get(publicationVersionId);
```
florin-holhos, line 19: The comment above can be misleading after this change. It should be removed or re-phrased.

ui/src/components/Publication/Visualization/index.tsx
```diff
@@ -17,7 +17,7 @@ interface BoxEntry {
     authorFirstName: string;
     authorLastName: string;
     publishedDate: string;
-    authors: Pick<Interfaces.CoAuthor, 'id' | 'linkedUser' | 'publicationId' | 'user'>[];
+    authors: Pick<Interfaces.CoAuthor, 'id' | 'linkedUser' | 'publicationVersionId' | 'user'>[];
```
florin-holhos, line 20: I have reworked Visualisation in #493 . Will need to handle some conflicts in this area either in that PR or in this one depending on which is merged first.

api/src/components/affiliations/controller.ts
finlay-jisc: Thanks, that's true. I just pushed a commit that rephrases this and a few similar comments elsewhere to be specific that they're dealing with versions.

florin-holhos (approved):

PR #489, opened 2023-09-21
OC-701: Maintenance mode

finlay-jisc:
The purpose of this PR was to add a way of stopping users from interacting with the application.

- I wanted to render this server side but it seems not to work unless the rest of the app is set up for this (see https://jiscdev.atlassian.net/browse/OC-700). Without the `mounted &&` I added, I get hydration errors to do with the ConfirmationModalProvider.
- Still need to confirm how it works on amplify.
- Hopefully the changes to header + footer are OK - maintenance mode was the only case I could think of where we would not want to show the parts of the header/footer which deal with the rest of octopus.

---

### Acceptance Criteria:

- Be toggleable easily, without deploying anything, if possible
- Not allow users to interact with the octopus front end
- Present a clear, user-friendly page explaining that the site is undergoing maintenance

---

### Checklist:

- [ ] Automated tests added
- [ ] Documentation updated

---

### Tests:

E2E:
Screenshot 2023-09-21 110817

UI tests are unavailable with the current dependencies (being worked on currently) but they passed when I downgraded nextJS to an older version:
Screenshot 2023-09-21 115252

---

### Screenshots:

Screenshot 2023-09-20 161415


ui/.env.example
```diff
@@ -5,4 +5,6 @@ NEXT_PUBLIC_BASE_URL=https://localhost:3001
 NEXT_PUBLIC_MEDIA_BUCKET=http://localhost:4566/science-octopus-publishing-images-local
 NEXT_PUBLIC_STAGE=local
 NEXT_PUBLIC_CONTENTFUL_SPACE_ID=contentful-space-id
-NEXT_PUBLIC_CONTENTFUL_ACCESS_TOKEN=contentful-access-token
\ No newline at end of file
+NEXT_PUBLIC_CONTENTFUL_ACCESS_TOKEN=contentful-access-token
+NEXT_PUBLIC_MAINTENANCE_MODE=false
+NEXT_PUBLIC_MAINTENANCE_MESSAGE=
```
florin-holhos, line 10: Is the second env var really needed?

florin-holhos (commented):
As discussed into the meeting, we just need to hide the top bar as well. Happy to approve once that's done

ui/.env.example
finlay-jisc: I double checked this with Tim and he said it could be useful to tell people timeframes + reasons for maintenance in future

finlay-jisc:
> As discussed into the meeting, we just need to hide the top bar as well. Happy to approve once that's done

I've set it to not render the header at all in the maintenance component so this should be OK now.

florin-holhos (approved):

PR #527, opened 2023-10-30
OC-389 Versioned DOIs

florin-holhos:
The purpose of this PR was to create a custom script in order to generate versioned DOIs and update existing (canonical) DOIs with the new generated DOIs for versions. Also on the API, when publishing, a new DOI will be generated for v1 and the relation between canonical DOI and v1 DOI will be updated on datacite as well.

For this specific task I had to duplicate /publications/[id] page and put it under /publications/[id]/versions/[versionId]
On prod users will still use the old publication page but on int and locally they will be redirected from /publications/[id] to /publications/[id]/versions/[versionId] until we develop further versioning functionalities and they are ready to go live on prod, but until then, individual tasks will be available for testing on int.

---

### Acceptance Criteria:

As per [OC-389](https://jiscdev.atlassian.net/browse/OC-389)

---

### Checklist:

- [ ] Automated tests added
- [ ] Documentation updated

---

### Tests:

---

### Screenshots:
image

image

[OC-389]: https://jiscdev.atlassian.net/browse/OC-389?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/prisma/createVersionedDOIs.ts
+++ b/api/prisma/createVersionedDOIs.ts
@@ -93,18 +93,17 @@ const createVersionedDOIs = async (): Promise<number> => {
 
         console.log(`Successfully created version ${version.versionNumber} DOI: ${versionDOI}`);
 
-        // update publication "HasVersion" with the created DOI
-        console.log(`Updating canonical DOI: ${version.publication.doi} "HasVersion" with version DOI: ${versionDOI}`);
-        await helpers.updatePublicationDOI(version.publication.doi, version);
-
-        console.log(`Successfully updated canonical DOI: ${version.publication.doi}`);
-
         // update version DOI
         console.log(`Updating version DOI: ${versionDOI} in DB`);
-
-        await publicationVersionService.update(version.id, {
+        const updatedVersion = await publicationVersionService.update(version.id, {
             doi: versionDOI
         });
+
+        // update publication "HasVersion" with the created DOI
+        console.log(`Updating canonical DOI: ${version.publication.doi} "HasVersion" with version DOI: ${versionDOI}`);
+        await helpers.updatePublicationDOI(version.publication.doi, updatedVersion);
+
+        console.log(`Successfully updated canonical DOI: ${version.publication.doi}`);
 
         createdVersionDOIsCount += 1;
         console.log(); // new line
--- a/api/prisma/createVersionedDOIs.ts
+++ b/api/prisma/createVersionedDOIs.ts
@@ -1,14 +1,16 @@
-import * as client from '../lib/client';
-import * as helpers from '../lib/helpers';
+import * as client from '../src/lib/client';
+import * as helpers from '../src/lib/helpers';
+import * as publicationVersionService from '../src/components/publicationVersion/service';
 
-/**
- * This script will update canonical DOIs with their latest versions
- */
-
-const updatePublicationDOIs = async (): Promise<void> => {
+const createVersionedDOIs = async (): Promise<number> => {
+    // get the latest LIVE version of each publication
     const latestLiveVersions = await client.prisma.publicationVersion.findMany({
         where: {
-            isLatestLiveVersion: true
+            isLatestLiveVersion: true,
+            createdBy: {
+                not: 'octopus' // ignore seed data publications
+            },
+            doi: null // only get versions without DOI
         },
         include: {
             publication: {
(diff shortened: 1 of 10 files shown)


api/prisma/createVersionedDOIs.ts
```diff
@@ -0,0 +1,120 @@
...
+        // create a new DOI for each version
+        console.log(`Creating version ${version.versionNumber} DOI for publication ${version.versionOf}`);
+        const versionDOIResponse = await helpers.createPublicationVersionDOI(version);
+        const versionDOI = versionDOIResponse.data.attributes.doi;
+
+        console.log(`Successfully created version ${version.versionNumber} DOI: ${versionDOI}`);
+
+        // update publication "HasVersion" with the created DOI
+        console.log(`Updating canonical DOI: ${version.publication.doi} "HasVersion" with version DOI: ${versionDOI}`);
+        await helpers.updatePublicationDOI(version.publication.doi, version);
```
finlay-jisc, line 98: Think there is a problem here - the second parameter we pass is the version we need, but it doesn't have the doi field set yet. So when it goes to set "HasVersion" in the helper function, the value of that ends up being null.

Screenshot 2023-11-01 131941

api/src/lib/helpers.ts
```diff
@@ -208,28 +378,29 @@ export const updateDOI = async (
...
-    const doiRes = await axios.put<I.DOIResponse>(`${process.env.DATACITE_ENDPOINT as string}/${doi}`, payload, {
+    const doi = await axios.post<I.DOIResponse>(`${process.env.DATACITE_ENDPOINT}`, payload, {
         auth: {
             username: process.env.DATACITE_USER as string,
             password: process.env.DATACITE_PASSWORD as string
         }
     });
 
-    return doiRes.data;
+    return doi.data;
```
finlay-jisc, line 403: The updatePublicationDOI and createPublicationVersionDOI functions are really similar. Can we just use one function and take a parameter to say whether we're dealing with a versionless / versioned DOI?

finlay-jisc (commented):
Although it's not on the jira item, I feel like I remember some discussion about the version DOIs to be of the form {doi base}/{versionless doi part}/{version number}, e.g.:

10.82259/cc8d-7421/1

With that we wouldn't have to save the version DOI in the DB because we can calculate it from the publication's ID and the version number.

I don't know if it's possible on the datacite end (maybe introducing another slash is invalid, or the ID is too long, or something) and it's not a big deal because this way works fine. If it wouldn't work or would be unnecessary effort to change there's no need to do anything.

florin-holhos:
> Although it's not on the jira item, I feel like I remember some discussion about the version DOIs to be of the form {doi base}/{versionless doi part}/{version number}, e.g.:
> 
> 10.82259/cc8d-7421/1
> 
> With that we wouldn't have to save the version DOI in the DB because we can calculate it from the publication's ID and the version number.
> 
> I don't know if it's possible on the datacite end (maybe introducing another slash is invalid, or the ID is too long, or something) and it's not a big deal because this way works fine. If it wouldn't work or would be unnecessary effort to change there's no need to do anything.

From what I read on the Datacite docs when working on this, generating DOIs of that form it's considered bad practice. There are some reserved characters that shouldn't be included into the suffix, including "/"
image

api/prisma/createVersionedDOIs.ts
florin-holhos: Yeah, I remember updating that helper function after I have created and tested the script already. Fixed it now.

api/src/lib/helpers.ts
florin-holhos: I will need to update these helpers in [OC-391](https://jiscdev.atlassian.net/browse/OC-391) anyway with "isPreviousVersionOf" and "isNewVersionOf" and also updating the canonical DOI when creating new version. I was thinking to create a separate transformer/mapper helper function/s which can be used by both of these helpers when creating/updating DOIs. For this specific use case, If we merge all the logic inside a single helper function we will end up with a huge function that's doing lots of unrelated stuff inside based on the extra params. We should split them up as much as we can and keep each helper function as simple as possible.

finlay-jisc: OK, sounds good also

api/prisma/createVersionedDOIs.ts
finlay-jisc: Looks good

finlay-jisc (approved):

PR #530, opened 2023-11-06
OC-721: Require login before checking whether a user has permission to view a draft

finlay-jisc:
The purpose of this PR was to redirect anonymous users to login when they attempt to view a publication that exists only in draft/locked status. This required a small API change so that we can determine from the API response status whether the publication exists in one of these states, or doesn't exist at all (which up to now both returned 404 status).

Also made a couple of end to end test tweaks as some of them were failing due to some recent versioning changes.

---

### Acceptance Criteria:

- When a user is not logged in and navigates directly to a draft publication, they should be redirected to the ORCiD log in, and directed back to the publication page.
  - If the user has permission to view it, they see the draft publication page as normal
  - If the user does not have permission to view it, they see a 404

---

### Checklist:

- [ ] Automated tests added
- [ ] Documentation updated

---

### Tests:

API:
Screenshot 2023-11-06 142613

UI:
Screenshot 2023-11-06 141912

E2E:
Screenshot 2023-11-06 142605

### DIFF
--- a/ui/src/lib/api.ts
+++ b/ui/src/lib/api.ts
@@ -1,4 +1,4 @@
-import axios, { AxiosResponse } from 'axios';
+import axios, { AxiosResponse, AxiosRequestConfig } from 'axios';
 
 import * as Interfaces from '@interfaces';
 import * as Config from '@config';
@@ -22,14 +22,29 @@ const api = axios.create({
     baseURL
 });
 
-export const get = async (url: string, token: string | undefined): Promise<AxiosResponse> => {
+export const get = async (
+    url: string,
+    token: string | undefined,
+    config?: AxiosRequestConfig
+): Promise<AxiosResponse> => {
     const headers = {
-        headers: {
-            Authorization: `Bearer ${token}`
-        }
+        Authorization: `Bearer ${token}`
     };
 
-    const response = await api.get(url, token ? headers : undefined);
+    const response = await api.get(
+        url,
+        config
+            ? {
+                  ...config,
+                  ...(token && { headers: {
+                    ...config.headers,
+                    ...headers
+                  } })
+              }
+            : token
+            ? { headers }
+            : undefined
+    );
     return response;
 };
 
--- a/ui/src/lib/api.ts
+++ b/ui/src/lib/api.ts
@@ -36,10 +36,12 @@ export const get = async (
         config
             ? {
                   ...config,
-                  ...(token && { headers: {
-                    ...config.headers,
-                    ...headers
-                  } })
+                  ...(token && {
+                      headers: {
+                          ...config.headers,
+                          ...headers
+                      }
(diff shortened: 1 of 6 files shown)


ui/src/pages/publications/[id]/index.tsx
```diff
@@ -58,15 +58,18 @@ export const getServerSideProps: Types.GetServerSideProps = async (context) => {
...
         Promise<Interfaces.Flag[] | void>,
         Promise<Interfaces.BaseTopic[] | void>
     ] = [
         api
-            .get(`${Config.endpoints.publications}/${requestedId}/publication-versions/latest`, token)
-            .then((res) => res.data)
+            .get(`${Config.endpoints.publications}/${requestedId}/publication-versions/latest`, token, {
+                // We need to know the return code of this request if it is 403/404 so we can redirect,
+                // so this stops it from being caught and not checkable in the following code.
+                validateStatus: (status) => [403, 404].includes(status) || (status >= 200 && status < 300)
```
Comment (florin-holhos) [inline ui/src/pages/publications/[id]/index.tsx:71]: I recommend using the .catch block and return the status code and error message from there if needed, instead of using "validateStatus".

ui/src/lib/api.ts
```diff
@@ -22,14 +22,31 @@ const api = axios.create({
     baseURL
 });
 
-export const get = async (url: string, token: string | undefined): Promise<AxiosResponse> => {
+export const get = async (
+    url: string,
+    token: string | undefined,
+    config?: AxiosRequestConfig
```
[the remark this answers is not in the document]
finlay-jisc: Yes, this makes sense. Just pushed code that uses the catch block instead.

ui/src/pages/publications/[id]/index.tsx
```diff
@@ -87,31 +96,47 @@ export const getServerSideProps: Types.GetServerSideProps = async (context) => {
...
+            return {
+                redirect: {
+                    destination: `${Config.urls.orcidLogin.path}&state=${encodeURIComponent(context.resolvedUrl)}`,
+                    permanent: false
+                }
+            };
+        }
+    }
+
+    if (publicationVersion) {
```
Comment (florin-holhos) [inline ui/src/pages/publications/[id]/index.tsx:122]: Do we need all these if/else checks? I think this part can be simplified like this:

if(!publicationVersion) {
    const status = versionRequestError?.status;
    
    return status === 403 ? { // redirect to login } : { // not found }
}

Comment (finlay-jisc) [reply ui/src/pages/publications/[id]/index.tsx:122]: I don't think we can simplify it functionally as we need to deal with 403s in two different ways depending on whether the user is logged in or not (404 if logged in, redirect to login if not).

We can change the if/else on line 108 to use ternary operator, but I don't think it will make much difference. To me, the if/else is slightly easier to read

florin-holhos (approved):

PR #532, opened 2023-11-09
OC-729: Vis not displaying correctly on a particular publication chain

finlay-jisc:
The purpose of this PR was to investigate and attempt to fix a couple of issues with the visualisation:

1. In some cases, the arrows cross over each other. This is partially fixed - the function that sorts the boxes in a given column only takes into account what is in the adjacent column, so can't fix more complex cases where it needs to know the rest of the chain (see 2nd screenshot).
2. When navigating through the chain to different publications, the arrows don't always re-render, and so can be left pointing in the wrong direction. This is fixed.

---

### Acceptance Criteria:

Ideally, stop these two issues from happening, or at least improve them.

---

### Checklist:

- [ ] Automated tests added
- [ ] Documentation updated

---

### Tests:

UI
Screenshot 2023-11-08 161307

E2E
Screenshot 2023-11-08 170628

---

### Screenshots:

This is an example using the same structure as the examples given in the item. This time, when viewing the interpretation, none of the arrows are crossing over.
Screenshot 2023-11-09 084848

However it's still possible to have more complex situations where the arrows can still cross over:
Screenshot 2023-11-09 085048

There will be some cases where this is unavoidable, so for now I would suggest we don't spend more effort catching the complex cases.


ui/src/components/Publication/Visualization/index.tsx
```diff
@@ -242,6 +207,137 @@ const Visualization: React.FC<VisualizationProps> = (props): React.ReactElement
...
+            if (publicationsOfType !== undefined) {
+                for (const publication of publicationsOfType) {
+                    // Get ids of children of this publication in the chain to which we will render a pointer arrow.
+                    const pointers =
+                        // This is the selected publication or its descendant, so get children from linkedFrom.
+                        type === selectedPublication.type ||
+                        filteredPublicationTypes.indexOf(type) >
+                            filteredPublicationTypes.indexOf(selectedPublication.type)
+                            ? [
+                                  ...new Set(
```
florin-holhos, line 299: Ordering the visualization boxes seems to add a fair amount of complexity with all these nested loops. This can have a negative impact on performance and ultimately affect users devices as the links count grows.

If possible (without too much hassle), I recommend having the links ordering logic (and any other hard work like this) on the backend and leave the UI as simple as possible so that it doesn't affect user's device performance.

finlay-jisc: I've just pushed a commit which moves the bulk of this to the publication service in the API

ui/src/components/Publication/Visualization/index.tsx
```diff
@@ -25,6 +26,10 @@ type BoxProps = BoxEntry & {
 };
 
 const Box: React.FC<BoxProps> = (props): React.ReactElement => {
+    const updateXarrow = useXarrow();
+    useEffect(() => {
+        updateXarrow();
+    }, [updateXarrow]);
```
florin-holhos, line 32: "updateXarrow" dependency is creating an infinite render loop and makes the page freeze
image

From what I see on "react-xarrows" documentation, we just need to call "useXarrow()" inside this component. We don't need the useEffect since "useXarrow" hook will be automatically called every time a visualization box is mounted into the DOM.
"updateXarrow" example is for other use cases like click or drag events.

ui/src/components/Publication/Visualization/index.tsx
```diff
@@ -242,6 +161,53 @@ const Visualization: React.FC<VisualizationProps> = (props): React.ReactElement
...
+                    title: publication.title,
+                    type: publication.type,
+                    createdBy: publication.createdBy,
+                    publishedDate: publication.publishedDate,
+                    authorFirstName: publication.authorFirstName,
+                    authorLastName: publication.authorLastName,
+                    authors: publication.authors,
+                    pointers:
+                        // for the selected publication and its child types, get pointers from linkedFrom
+                        publication.type === selectedPublication.type || typeIndex > selectedTypeIndex
```
florin-holhos, line 183: Is this code (from line 165 to line 209) supposed to do the same thing as "getPublicationsByType" helper function which has been removed? If so, I strongly suggest keeping that function instead of all this code because it was way more human readable, easier to understand and maintain and also better optimised.

ui/src/components/Publication/Visualization/index.tsx
finlay-jisc: I see, I wasn't too sure about that one, that makes sense. Just pushed the fix

ui/src/components/Publication/Visualization/index.tsx
finlay-jisc: It's meant to do basically the same thing, I just thought it would be simpler to do it this way by looping through each publication and making one box for each. With the getPublicationsByType function, for every type, we're looping through at least linkedTo or linkedFrom (or both, if the selected publication is a problem or real world application); that seems like more looping to me. Also the way the pointers are set is a little bit different.

I'm happy to tidy this up more to make it more readable - it would be good to know what's making it less readable so I can change that. Or if I've missed something about optimisation, what that is so I can change that too.

florin-holhos: With the getPublicationsByType function, for every type, we're looping through and get all pubs of that type and render them as a column, simple and easy to follow. It has a maximum complexity of O(n^2) and also while rendering it has the same complexity because there's only a "map" in "map" which is not great either but still better than "map" in ".filter(...).map(...)" 

With this change, from line 166 through line 189 there's a O(n^3) complexity because there's a `for > ...spreadOperator > .filter/.map` + an extra unnecessary O(n^2) time complexity with that .indexOf inside the for(...) loop. When rendering the boxes, again (as I've mentioned above), there's an extra .filter() while mapping  `.map > .filter > .map again`. The way the pointers are set is harder to read and understand than before with all these conditions, ternary operators and spreads.

After all, this change is not needed and I don't see it as an improvement since it's doing the same thing but adding more complexity and it's harder to read. New devs will just spend more time understanding it.

florin-holhos (approved):

PR #537, opened 2023-11-24
OC-731: Re-designing the account page to accommodate multiple versions

finlay-jisc:
The purpose of this PR was to clarify the "My Account" page by showing 1 row per publication and clearly showing whether it has a draft and/or published version.

---

### Acceptance Criteria:
As per [OC-392](https://jiscdev.atlassian.net/browse/OC-392)

---

### Checklist:

- [x] Automated tests added
- [ ] Documentation updated

---

### Tests:

API:
Screenshot 2023-11-24 104609

UI:
Screenshot 2023-11-24 104628

E2E:
Screenshot 2023-11-24 103241

[OC-392]: https://jiscdev.atlassian.net/browse/OC-392?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ui/src/components/Button/index.tsx
```diff
@@ -12,6 +12,7 @@ type CommonProps = {
     textSize?: string;
     padding?: string;
     children?: React.ReactNode;
+    childClasses?: string;
```
florin-holhos, line 15: There's a `children:` variant for this. You can use it in the className prop like this: `children:border-none` in order to style the button child nodes.

ui/src/components/Publication/SimpleResult/index.tsx
```diff
@@ -19,142 +19,169 @@ const SimpleResult: React.FC<Props> = (props): React.ReactElement => {
...
+                                href={`/publications/${props.publication.id}`}
+                                endIcon={<OutlineIcons.EyeIcon className="h-4" />}
+                                title="View Draft"
+                                className="mt-5 bg-green-600 px-3 text-white-50"
+                                childClasses="text-white-50 border-none"
+                            />
+                        )}
+                    </>
+                ) : authorOnLatestLive ? (
+                    draftExistsWithoutPermission ? (
```
florin-holhos, line 138:
```suggestion
                ) : isAuthorOnLatestLive ? (
                    draftExistsWithoutPermission ? (
```

ui/src/components/Publication/SimpleResult/index.tsx
```diff
@@ -19,142 +19,169 @@ const SimpleResult: React.FC<Props> = (props): React.ReactElement => {
...
+                        <div className="flex-grow">
+                            <p>
+                                Published on:{' '}
+                                <time suppressHydrationWarning>
+                                    {latestLiveVersion.publishedDate &&
+                                        Helpers.formatDate(latestLiveVersion.publishedDate)}
+                                </time>
+                            </p>
+                            {!authorOnLatestLive && <p>You are not listed as an author on the latest version</p>}
+                        </div>
```
florin-holhos, line 171:
```suggestion
                            {!isAuthorOnLatestLive && <p>You are not listed as an author on the latest version</p>}
                        </div>
```

ui/src/components/Publication/SimpleResult/index.tsx
```diff
@@ -19,142 +19,169 @@ const SimpleResult: React.FC<Props> = (props): React.ReactElement => {
...
-    );
+
+    const latestLiveVersion = props.publication.versions.find((version) => version.isLatestLiveVersion);
+    const authorOnLatestLive =
+        latestLiveVersion?.coAuthors.find((coAuthor) => coAuthor.linkedUser === props.user.id) ||
+        latestLiveVersion?.createdBy === props.user.id;
+    const latestVersion = props.publication.versions.find((version) => version.isLatestVersion);
+    const draftExistsWithPermission =
+        latestVersion?.currentStatus === 'DRAFT' || latestVersion?.currentStatus === 'LOCKED';
+    const draftVersion = draftExistsWithPermission ? latestVersion : false;
```
florin-holhos, line 30: Using `latestVersion : false` can be misleading. I guess we can replace `false` with `null` or `undefined` if latestVersion does not exists.

ui/src/components/Button/index.tsx
finlay-jisc: That's handy, I have switched it to use that

ui/src/components/Publication/SimpleResult/index.tsx
```diff
@@ -19,142 +19,169 @@ const SimpleResult: React.FC<Props> = (props): React.ReactElement => {
...
     const [error, setError] = React.useState('');
     const [loading, setLoading] = React.useState(false);
-    const status = useMemo(
-        () => Helpers.getPublicationStatusByAuthor(props.publicationVersion, props.user),
-        [props.publicationVersion, props.user]
-    );
+
+    const latestLiveVersion = props.publication.versions.find((version) => version.isLatestLiveVersion);
+    const authorOnLatestLive =
+        latestLiveVersion?.coAuthors.find((coAuthor) => coAuthor.linkedUser === props.user.id) ||
```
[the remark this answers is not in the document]
finlay-jisc: Yeah, that would be clearer. I have committed it

ui/src/components/Publication/SimpleResult/index.tsx
```diff
@@ -19,142 +19,165 @@ const SimpleResult: React.FC<Props> = (props): React.ReactElement => {
     const router = useRouter();
     const [error, setError] = React.useState('');
```
florin-holhos, line 20: I see this error state is not being used anymore, but it's still being set from try/catch blocks. What happens if the user clicks on 'Create Draft Version' and the API returns an error? Also, if this state it's not needed, we should remove it completely.

ui/src/components/Publication/SimpleResult/index.tsx
finlay-jisc: Changed this to use null instead of false.

ui/src/components/Publication/SimpleResult/index.tsx
finlay-jisc: Doesn't look like it was ever specced, but it makes sense to show an error message if the new draft creation fails. I've added an alert in that case

ui/next.config.js
```diff
@@ -4,13 +4,18 @@ const nextConfig = {
...
-          },
-        ]
-      },
+            {
+                source: '/search',
+                destination: '/search/publications/',
+                permanent: true
+            },
+            {
+                source: '/publications/:path',
```
florin-holhos, line 13: This type of redirect is ok for server-side redirects, eg: when the user tries to access a publication URL directly.
But, it's not working properly with client-side redirects using next/router and next/link. All the links and other client-side redirects like `router.push(...)` pointing to `/publications/${id}` will trigger a hard page reload because of this redirect. While redirecting, the page is blank for a few seconds and it's loading very very slow. This a known problem with Next.js when using redirects, which unfortunately won't be fixed any time soon. I've came into this problem when I added the redirect from /publications to /publications/id/versions/latest.

I see two potential fixes for this.

1. To remove this redirect and leave the publication page + redirect from getServerSideProps as it was because redirecting from getServerSideProps works fine, without hard reloads (which is also interesting, but... it's Next.js...). The publication page can be blank, it doesn't need to fetch or render anything, it just needs to return a redirect from getServerSideProps.

2. to keep this redirect for server-side redirects and update all links and redirects which currently point to /publications/{id} to point directly to the latest version of that publication /publication/{id}/versions/latest

finlay-jisc: Thanks for the explanation - I thought fix 1 would be the best so I've moved it over to getServerSideProps in the latest commit.

ui/src/pages/publications/[id]/index.tsx
```diff
@@ -1,929 +1,16 @@
...
-                );
-
-                setBookmarkId(newBookmarkResponse.data?.id);
-            } catch (err) {
-                console.log(err);
-            }
+    return {
+        redirect: {
+            destination: `/publications/${requestedId}/versions/latest`, // this url might change in OC-391
+            permanent: false
```
Comment (florin-holhos) [inline ui/src/pages/publications/[id]/index.tsx:9]: This is working fine now. The `permanent` should be set to `true` now since we release this feature and the redirect is not temporary anymore.

ui/src/pages/publications/[id]/index.tsx
```diff
@@ -1,929 +1,16 @@
...
-                    user?.token
-                );
-
-                setBookmarkId(newBookmarkResponse.data?.id);
-            } catch (err) {
-                console.log(err);
-            }
+    return {
+        redirect: {
+            destination: `/publications/${requestedId}/versions/latest`, // this url might change in OC-391
```
Comment (florin-holhos) [inline ui/src/pages/publications/[id]/index.tsx:8]: I guess we can remove the comment since that ticket is done

florin-holhos (approved):

PR #553, opened 2023-12-18
OC-745: Generate separate sitemaps per 50k publications on back end

finlay-jisc:
The purpose of this PR was to create a sitemap to aid the indexing of our pages by search engines. In order to be future proof (for when we have more content), this should consist of separate sitemaps for every 50000 URLs we intend to map. To do this we will create multiple sitemaps for every 50000 live publications. Other content (static pages, topics, blog posts) are covered in a single standalone sitemap.

---

### Acceptance Criteria:

- Every 24 hours, a cron job triggers a lambda to generate a sitemap for publications which is then stored in an s3 bucket 
- An API endpoint is present which returns links to XML sitemaps for publications
- The sitemaps at /publications/[number].xml in an s3 bucket contain links to the “[number]th” 50,000 publications in octopus (a consistent sort order between the sitemaps is needed but it doesn’t especially matter what it is). For example, /publications/3.xml would contain the 100,001st publication up to the 150,000th publication.

---

### Checklist:

- [ ] Automated tests added
- [ ] Documentation updated


api/prisma/seed.ts
```diff
@@ -101,49 +101,35 @@ export const initialDevSeed = async (): Promise<void> => {
...
-        } catch (err) {
-            // Bucket does not exist, therefor create
-            await s3.client.send(
-                new CreateBucketCommand({
-                    Bucket: `science-octopus-publishing-pdfs-${process.env.STAGE}`
-                })
-            );
-            console.log('Bucket created');
+        // Create local S3 buckets
+        for (const bucketNameSegment of ['images', 'pdfs', 'sitemaps']) {
```
florin-holhos, line 105: You can do `for (const bucketNameSegment in s3.buckets) { ...` instead of the hardcoded array

api/serverless.yml
```diff
@@ -548,3 +548,17 @@ functions:
         handler: src/components/event/routes.dailyEventCheck
         events:
             - schedule: cron(0 8 * * ? *) # Every day at 8 a.m.
+
+    # Sitemaps
+    generateSitemaps:
+        handler: src/components/sitemap/controller.generate
```
florin-holhos, line 554: There's a route.ts file missing under /components/sitemap folder. All the other functions are using (or at least they should use) the methods exported from routes.ts and not from controller.ts directly, for consistency.

api/src/components/sitemap/controller.ts
```diff
@@ -0,0 +1,33 @@
+import * as I from 'interface';
+import * as response from 'lib/response';
+import * as sitemapService from 'sitemap/service';
+
+export const generate = async (): Promise<I.JSONResponse> => {
+    try {
+        const sitemapCount = await sitemapService.generatePublicationSitemaps();
+
+        return response.json(200, {
```
florin-holhos, line 9: This is not an API endpoint, it doesn't need to return anything

api/src/lib/s3.ts
```diff
@@ -17,4 +17,15 @@ if (process.env.STAGE === 'local') {
     };
 }
 
+type OctopusBucketType = 'images' | 'pdfs' | 'sitemaps';
+const buckets: {
+    [key in OctopusBucketType]?: string;
+} = {};
+
+for (const nameSegment of ['images', 'pdfs', 'sitemaps']) {
```
florin-holhos, line 25: Not a huge fan of hardcoded array values like these without any type checking. 
nameSegment could be anything...

You can convert type OctopusBucketType to enum and then easily loop through it or simply remove the optional key in buckets and define the bucket for each key there, there are just 3 buckets.

ui/src/pages/sitemaps/other.xml.tsx
```diff
@@ -0,0 +1,130 @@
...
+                    'authors',
+                    'login.tsx',
+                    'my-bookmarks.tsx',
+                    'publications',
+                    'publications-sitemap-index.xml.tsx',
+                    'publications-sitemap[number].xml.tsx',
+                    'sitemap.xml.tsx',
+                    'topics',
+                    'verify.tsx'
+                ].includes(staticPage)
```
florin-holhos, line 38: We may need to refine this filter a bit to also catch paths starting with period ".", "approve-control-request.tsx"
image

api/prisma/seed.ts
finlay-jisc: That's better, thanks, done

api/serverless.yml
finlay-jisc: Ok, have set this up

api/src/components/sitemap/controller.ts
finlay-jisc: Have tidied up the sitemap controller + service along these lines

api/src/lib/s3.ts
finlay-jisc: Yeah, my initial idea to loop through since most of the name is repeated for each bucket ended up being more mess than it's worth. I've ended up defining it simply as you suggested

ui/src/pages/sitemaps/other.xml.tsx
finlay-jisc: I've adjusted this list so that it will exclude files starting with '.' and 'approve-control-request.tsx'

ui/src/pages/sitemap.xml.tsx
```diff
@@ -0,0 +1,45 @@
+import * as Types from '@types';
+import * as api from '@api';
+import * as Config from '@config';
+
+const PublicationSitemapIndex = () => {
```
florin-holhos, line 5: The name should be Sitemap. This should return the sitemap index with all our sitemaps located under /sitemaps

Under /sitemaps folder we do all the fetching and filtering if needed. Each resource that needs to be fetched in order to generate a sitemap index should be located under /sitemaps for example:

`/sitemaps
            /static.xml.tsx
            /publications.xml.tsx
            /topics.xml.tsx
            /users.xml.tsx
            /blogs.xml.tsx`

ui/src/pages/sitemaps/other.xml.tsx
```diff
@@ -0,0 +1,130 @@
...
+                        <url>
+                            <loc>${url}</loc>
+                            <lastmod>${now}</lastmod>
+                            <changefreq>monthly</changefreq>
+                            <priority>1.0</priority>
+                        </url>
+                    `
+                )
+                .join('')}
+            ${topicUrls
```
florin-holhos, line 93: From this file we only need to return the sitemap with the static pages. The other resources like topics and blogs can be handled in their own file

ui/src/pages/sitemap.xml.tsx
finlay-jisc: I've split these out into seperate files in the latest commit; they each export a var called Sitemap

ui/src/pages/sitemaps/other.xml.tsx
finlay-jisc: I've restructured the files in the latest commit

florin-holhos (approved):

adwearing-jisc (approved):

PR #554, opened 2023-12-20
OC-742: Ensure all edits affect "Last Updated" on account page

finlay-jisc:
The purpose of this PR was to make sure that all actions a user would understand to be "updating a publication (version)" update the updatedAt field on the publication version model in the database. Then when this is surfaced in the application it will match up to the actions they have taken.

---

### Acceptance Criteria:

The following items should affect the last updated date of publications on the account page:

1. Edits to the publication version
2. All related entities
3. Funder information
4. Affiliations changes (co author and corresponding author)
5. Co-author approval
6. Co-author changing their mind
7. Co-author denying involvement
8. Co-author being added or removed
9. Publication being locked or unlocked
10. Transfers of corresponding authorship

---

### Checklist:

- [ ] Automated tests added
- [ ] Documentation updated

---

### Tests:

API:
Screenshot 2023-12-20 094900

E2E:
Screenshot 2023-12-20 094202

### DIFF
--- a/api/src/components/coauthor/service.ts
+++ b/api/src/components/coauthor/service.ts
@@ -1,5 +1,6 @@
 import * as client from 'lib/client';
 import * as I from 'lib/interface';
+import * as publicationVersionService from 'publicationVersion/service';
 import { createId } from '@paralleldrive/cuid2';
 import { Prisma } from '@prisma/client';
 
@@ -48,13 +49,8 @@ export const update = async (id: string, data: Prisma.CoAuthorsUpdateInput) => {
         data
     });
 
-    await client.prisma.publicationVersion.update({
-        where: {
-            id: updateCoAuthor.publicationVersionId
-        },
-        data: {
-            updatedAt: new Date().toISOString()
-        }
+    await publicationVersionService.update(updateCoAuthor.publicationVersionId, {
+        updatedAt: new Date().toISOString()
     });
 
     return updateCoAuthor;
@@ -89,13 +85,8 @@ export const updateAll = async (publicationVersionId: string, authors: I.CoAutho
         )
     );
 
-    await client.prisma.publicationVersion.update({
-        where: {
-            id: publicationVersionId
-        },
-        data: {
-            updatedAt: new Date().toISOString()
-        }
+    await publicationVersionService.update(publicationVersionId, {
+        updatedAt: new Date().toISOString()
     });
 
     return update;
@@ -115,14 +106,12 @@ export const deleteCoAuthor = async (id: string) => {
             id
         }
     });
-    await client.prisma.publicationVersion.update({
-        where: {
-            id: coAuthor?.publicationVersionId
-        },
-        data: {
+
+    if (coAuthor) {
+        await publicationVersionService.update(coAuthor.publicationVersionId, {
             updatedAt: new Date().toISOString()
-        }
-    });
+        });
+    }
 
     return deleteCoAuthor;
 };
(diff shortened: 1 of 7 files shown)


api/src/components/coauthor/service.ts
```diff
@@ -40,14 +40,26 @@ export const getAllByPublicationVersion = async (publicationVersionId: string) =
...
-    client.prisma.coAuthors.update({
+export const update = async (id: string, data: Prisma.CoAuthorsUpdateInput) => {
+    const updateCoAuthor = await client.prisma.coAuthors.update({
         where: {
             id
         },
         data
     });
 
+    await client.prisma.publicationVersion.update({
```
florin-holhos, line 51: We should use the update method exported from the publicationVersion/service.ts for all these updates

finlay-jisc: Just pushed a commit that converted these calls to use the service

api/src/components/link/service.ts
```diff
@@ -8,6 +9,20 @@ export const create = async (fromPublicationId: string, toPublicationId: string)
         }
     });
 
+    const latestVersionFrom = await client.prisma.publicationVersion.findFirst({
```
florin-holhos, line 12: We already have a service for this:
`const latestVersionFrom = await publicationVersionService.get(fromPublicationId, 'latest');`

api/src/components/link/service.ts
```diff
@@ -8,6 +9,20 @@ export const create = async (fromPublicationId: string, toPublicationId: string)
...
+        where: {
+            versionOf: fromPublicationId,
+            isLatestVersion: true
+        },
+        select: { id: true }
+    });
```
florin-holhos, line 18:
```suggestion
const latestVersionFrom = await publicationVersionService.get(fromPublicationId, 'latest');
```

api/src/components/link/service.ts
```diff
@@ -23,12 +38,34 @@ export const doesLinkExist = async (fromPublicationId: string, toPublicationId:
...
+                select: {
+                    id: true
+                }
+            }
+        }
+    });
+    const latestVersionFrom = await client.prisma.publicationVersion.findFirst({
+        where: { versionOf: link?.publicationFromRef.id, isLatestVersion: true },
+        select: { id: true }
+    });
```
florin-holhos, line 56:
```suggestion
const latestVersionFrom = await publicationVersionService.get(link?.publicationFromRef.id, 'latest');
```

api/src/components/link/service.ts
```diff
@@ -23,12 +38,34 @@ export const doesLinkExist = async (fromPublicationId: string, toPublicationId:
...
+        },
+        select: {
+            publicationFromRef: {
+                select: {
+                    id: true
+                }
+            }
+        }
+    });
+    const latestVersionFrom = await client.prisma.publicationVersion.findFirst({
```
florin-holhos, line 53: The same as above, we already have a service which returns the latest version for a given publication id

api/src/components/link/service.ts
finlay-jisc: I tried to use this initially but typescript didn't think the id field was available on the result

florin-holhos: It works... you may need an if statement in some cases to make TS happy but in this case there's an if statement below this
image

finlay-jisc: Yeah it works... I don't know what was different before. Now pushed

florin-holhos (approved):

PR #563, opened 2024-01-03
OC-762: Co-authors unable to view locked drafts

finlay-jisc:
The purpose of this PR was to show a View Draft button for publications on the account page when the user is a confirmed coauthor on a locked publication version.

Unfortunately some of our tests just run over 3 minutes now so I increased the timeout further. We should really restructure these soon to make them much quicker because it's causing small changes to take much longer than they need to if the full tests are run. Then this timeout can be reduced to a sensible value.

---

### Acceptance Criteria:

- As a co-author, whilst a publication is displaying the “pending your approval” status, and is locked, I see a “View Draft” button

---

### Checklist:

- [ ] Automated tests added
- [ ] Documentation updated

---

### Screenshots:

Screenshot 2024-01-02 153000

### DIFF
--- a/e2e/playwright.config.ts
+++ b/e2e/playwright.config.ts
@@ -14,7 +14,7 @@ dotenv.config();
 const config: PlaywrightTestConfig = {
     testDir: './tests',
     /* Maximum time one test can run for. */
-    timeout: 180000, // some of the publication flow and coauthor tests exceed 2 minutes. We should try and streamline them but for now set this to 3 mins
+    timeout: 300000, // some of the publication flow and coauthor tests exceed 2 minutes. We should try and streamline them but for now set this to 3 mins
     expect: {
         /**
          * Maximum time expect() should wait for the condition to be met.
--- a/e2e/tests/LoggedIn/publish.e2e.spec.ts
+++ b/e2e/tests/LoggedIn/publish.e2e.spec.ts
@@ -796,6 +796,7 @@ const checkPublicationOnAccountPage = async (
     if (navigate) {
         await page.locator(PageModel.header.usernameButton).click();
         await page.locator(PageModel.header.myAccountButton).click();
+        await page.waitForURL(`${Helpers.UI_BASE}/account`);
     }
     const publicationContainer = await page.getByTestId('publication-' + publicationDetails.id);
     switch (state) {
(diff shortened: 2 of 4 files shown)


e2e/tests/LoggedIn/publish.e2e.spec.ts
```diff
@@ -815,6 +816,7 @@ const checkPublicationOnAccountPage = async (
         case 'pending your approval':
             await expect(publicationContainer).toContainText('(Author)');
             await expect(publicationContainer).toContainText('Status: Pending your approval');
+            await expect(publicationContainer.locator(PageModel.myAccount.viewDraftButton)).toBeVisible();
```
adwearing-jisc, line 819: Is it worth having a test to confirm this doesn't show when in other states? (I think just the approved state).

e2e/tests/LoggedIn/publish.e2e.spec.ts
```diff
@@ -837,12 +839,14 @@ const checkPublicationOnAccountPage = async (
             await expect(publicationContainer).toContainText(
                 'Someone else is working on a new draft version, and you do not yet have access to it'
             );
+            await expect(publicationContainer.locator(PageModel.myAccount.requestControlButton)).toBeVisible();
```
adwearing-jisc, line 842: Same as above comment, could be helpful to have a test to confirm this one doesn't show on other states

e2e/playwright.config.ts
```diff
@@ -14,7 +14,7 @@ dotenv.config();
 const config: PlaywrightTestConfig = {
     testDir: './tests',
     /* Maximum time one test can run for. */
-    timeout: 180000, // some of the publication flow and coauthor tests exceed 2 minutes. We should try and streamline them but for now set this to 3 mins
+    timeout: 300000, // some of the publication flow and coauthor tests exceed 2 minutes. We should try and streamline them but for now set this to 3 mins
```
adwearing-jisc, line 17: 5 minutes for a test 😨

e2e/tests/LoggedIn/publish.e2e.spec.ts
finlay-jisc: Actually we do want to show it in approved state so that the author can go to the draft and publish it as the next step, but have pushed a commit to be more specific about when the view draft button should and shouldn't show.

e2e/tests/LoggedIn/publish.e2e.spec.ts
finlay-jisc: Agreed, have pushed a commit that specifies the scenarios when viewing as a coauthor in which this button should not be shown.

e2e/playwright.config.ts
finlay-jisc: It's bad :( but we have an item this sprint to begin streamlining these tests so that they are more reliable and quick to run (OC-764).

adwearing-jisc (approved):

PR #580, opened 2024-01-29
OC-746: Generate separate sitemaps per 50k users on back end

finlay-jisc:
The purpose of this PR was to have sitemaps of a size that google can manage. We need to make sure that they only contain 50,000 URLs. In the case of users, we should develop a solution that can cope with the system having over 50,000 users.

---

### Acceptance Criteria:

- Every 24 hours, a cron job triggers a lambda to generate a sitemap for users which is then stored in an s3 bucket 
- The API endpoint which is used to return XML sitemaps for publications (OC-745) also returns the sitemaps for users
- The sitemaps at /users/[number].xml in an s3 bucket contain links to the “[number]th” 50,000 user in octopus (a consistent sort order between the sitemaps is needed but it doesn’t especially matter what it is). For example, /sitemaps/users/3.xml would contain the 100,001st user up to the 150,000th user.

---

### Checklist:

- [ ] Automated tests added
- [ ] Documentation updated

---

### Screenshots:

Screenshot 2024-01-29 093454

Screenshot 2024-01-29 093509

### DIFF
--- a/api/src/components/sitemap/service.ts
+++ b/api/src/components/sitemap/service.ts
@@ -42,7 +42,7 @@ export const generateSitemaps = async (category: 'publications' | 'users'): Prom
     // Get IDs for first sitemap
     let queryResults =
         category === 'publications'
-            ? await client.prisma.publication.findMany(queryOptions )
+            ? await client.prisma.publication.findMany(queryOptions)
             : await client.prisma.user.findMany(queryOptions as Prisma.UserFindManyArgs);
     let sitemapCount = 0;
 
@@ -81,7 +81,7 @@ export const generateSitemaps = async (category: 'publications' | 'users'): Prom
                       cursor: {
                           id: queryResults[queryResults.length - 1].id
                       },
-                      ...(queryOptions )
+                      ...queryOptions
                   })
                 : await client.prisma.user.findMany({
                       skip: 1,
--- a/api/src/components/sitemap/service.ts
+++ b/api/src/components/sitemap/service.ts
@@ -16,19 +16,21 @@ const listSitemapBucket = async (): Promise<ListObjectsV2CommandOutput> => {
     );
 };
 
-export const generatePublicationSitemaps = async (): Promise<void> => {
+export const generateSitemaps = async (category: 'publications' | 'users'): Promise<void> => {
     // The maximum number of URLs per sitemap (the maximum google can take in one go)
     const URL_LIMIT = 50000;
-    const now = new Date().toISOString();
-    const queryOptions: Prisma.PublicationFindManyArgs = {
+    // DB query options - publications should only be live ones, users can be any.
+    const queryOptions: Prisma.PublicationFindManyArgs | Prisma.UserFindManyArgs = {
         take: URL_LIMIT,
-        where: {
-            versions: {
-                some: {
-                    currentStatus: 'LIVE'
+        ...(category === 'publications' && {
+            where: {
+                versions: {
+                    some: {
+                        currentStatus: 'LIVE'
+                    }
                 }
             }
-        },
+        }),
         select: {
             id: true
         },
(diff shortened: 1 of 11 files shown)


api/src/components/sitemap/service.ts
```diff
@@ -62,32 +69,42 @@ export const generatePublicationSitemaps = async (): Promise<void> => {
...
             })
         );
-        queryResults = await client.prisma.publication.findMany({
-            skip: 1,
-            cursor: {
-                id: queryResults[queryResults.length - 1].id
-            },
-            ...queryOptions
-        });
+        queryResults =
```
adwearing-jisc, line 77: I know this is slightly out of scope for this PR, but this method of getting the publications/users is a bit odd.
Get 50k, do something with them, then get the next 50k, do something with them etc etc.

Would it make more sense to get the lot, split into 50k buckets and dispatch the buckets off to google one by one? I've done a similar thing here - https://github.com/JiscSD/github-monitor/blob/d9221383d80ef52b26ae6710d5fecd5adf9a6cd1/src/utils/DynamoStore.ts#L14 - gets all data, splits into 25 item chunks (dynamodb limit) then loops through chunks saving them to database.

Only potential issue could be the sheer amount of sitemaps coming from database..

ui/src/lib/helpers.tsx
```diff
@@ -593,3 +593,27 @@ export const toKebabCase = (inputString: string): string => {
         .replace(/\s+/g, ' ') // Condense longer whitespace down to one space
         .replace(/\s/g, '-'); // Replace single spaces with hyphen
 };
+
+// Fetches sitemaps of a category from the sitemap S3 bucket and constructs a sitemap index of them.
+export const getSitemapIndexXML = async (category: 'publications' | 'users'): Promise<string> => {
+    let sitemapS3Keys: string[] = [];
+    try {
+        const sitemapsRequest = await api.get('/sitemaps/paths', undefined);
```
adwearing-jisc, line 601: Should be able to omit the second parameter, undefined would be default if not set

ui/src/lib/helpers.tsx
```diff
@@ -593,3 +593,27 @@ export const toKebabCase = (inputString: string): string => {
...
+        sitemapS3Keys = sitemapsRequest.data;
+    } catch (error) {
+        console.log(error);
+    }
+    // Write an XML element for each applicable sitemap we have in S3
+    const sitemapXMLElements = sitemapS3Keys
```
adwearing-jisc, line 607: Using .filter might be cleaner and easier to understand here

```
sitemapS3Keys.filter(k => k.startsWith(`${category}/`)).map(k => `<sitemap...></sitemap>`).join('');
```

This way you aren't having to use flatmap to flatten either an array or a string (which is confusing) and the filtering is a distinct step rather than being mixed in with the mapping.

adwearing-jisc: Although I should probably ask how many sitemapS3Keys are there likely to be? Might not be ideal doing separate filter step if there's a large amount due to looping through array twice.

ui/src/pages/sitemaps/users.xml.tsx
```diff
@@ -0,0 +1,20 @@
+import * as Helpers from '@/helpers';
```
adwearing-jisc, line 1: Why .xml in the filename? Does that do something different in nextjs?

adwearing-jisc (approved):

ui/src/lib/helpers.tsx
finlay-jisc: The type signature for api.get required a 2nd param; I've tweaked this so that we don't have to pass a 2nd parameter anymore

ui/src/lib/helpers.tsx
finlay-jisc: This is done now!

ui/src/pages/sitemaps/users.xml.tsx
finlay-jisc: Yeah, this is just to control the eventual path - if it didn't contain ".xml" the eventual path would be /sitemaps/users, so instead we've chosen to be explicit about the file type that's going to be delivered.

api/src/components/sitemap/service.ts
adwearing-jisc: Yeah either way works I guess. My preference is always to limit the amount of round trips to database, but tradeoff is inefficiency of javascript. I guess my example doesn't have >50k entries to process 🙃

adwearing-jisc (approved):

PR #596, opened 2024-02-07
OC-767: Sending new versions to pubrouter

finlay-jisc:
The purpose of this PR was to send an updated record to publications router under the “all versions” DOI when a new version of a publication is published.

Also moved the generatePDF function from the publication service to the publication version service as per the TODO comment.

Some parts of this flow are unavailable locally so the full flow including this part will have to be confirmed on the int environment:
- Triggering PDF generation via SQS messages
    - Locally, they're generated when requested, and this won't force a regenerate like the SQS triggered calls will
- Triggering the message to pubrouter when a PDF is put in the S3 bucket

Also fleshed out a basic todo e2e test to confirm that PDFs are generated.

---

### Acceptance Criteria:

Upon the publication of a new version, a new PDF for that version is generated, triggering a new notification to publications router containing all details about the new version, using the “all versions”/canonical DOI.

---

### Checklist:

- [x] Automated tests added
- [ ] Documentation updated

---

### Tests:

UI
Screenshot 2024-02-07 134247

API
Screenshot 2024-02-07 133813

E2E
Screenshot 2024-02-07 133125

### DIFF
--- a/e2e/tests/LoggedIn/livePublication.e2e.spec.ts
+++ b/e2e/tests/LoggedIn/livePublication.e2e.spec.ts
@@ -24,7 +24,10 @@ export const testBookmarking = async (page: Page, id: string) => {
     await page.locator(PageModel.header.myBookmarksButton).click();
 
     await page.waitForSelector(PageModel.myBookmarks.publicationBookmark);
-    await expect(page.locator(PageModel.myBookmarks.publicationBookmark)).toHaveAttribute('href', `/publications/${id}`);
+    await expect(page.locator(PageModel.myBookmarks.publicationBookmark)).toHaveAttribute(
+        'href',
+        `/publications/${id}`
+    );
 
     // Remove bookmark
     await page.locator(PageModel.myBookmarks.publicationBookmark).click();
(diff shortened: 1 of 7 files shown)


e2e/tests/LoggedIn/livePublication.e2e.spec.ts
```diff
@@ -127,16 +130,28 @@ test.describe('Live Publication', () => {
...
         await Helpers.login(page, browser);
         await expect(page.locator(PageModel.header.usernameButton)).toHaveText(`${Helpers.user1.fullName}`);
         await page.goto(`${Helpers.UI_BASE}/publications/cl3fz14dr0001es6i5ji51rq4`, { waitUntil: 'domcontentloaded' });
+
+        // Behaviour is different depending on whether test is running headless or not.
+        if (headless) {
+            // Download PDF
+            const downloadPromise = page.waitForEvent('download');
+            await page.locator('aside').getByLabel('Download PDF').click();
+            const download = await downloadPromise;
```
adwearing-jisc, line 145: Curious, in my experience with these kind of events I've needed to await both promises at the same time in a promise.all, otherwise there's a chance the download event fires before you start awaiting the download event.

Is this showing flaky behaviour at all?

adwearing-jisc: Of course playwright are being inconsistent here.. Elsewhere they recommend doing a promise.all when waiting for events.

It might be ok then if playwright suggest it this way.

PR #599, opened 2024-02-13
OC-805: Author search not responding to search terms

finlay-jisc:
The purpose of this PR was to fix a bug where entering a query in the "quick search" box on the author search page didn't affect the results.

Added some tests for our PageTitle component.

---

### Acceptance Criteria:

- When a search term is submitted into the author search, results are filtered to match the term against the author's first or last name.

---

### Checklist:

- [x] Automated tests added
- [ ] Documentation updated

---

### Tests:

UI
Screenshot 2024-02-13 102603

E2E
Screenshot 2024-02-13 102540

---

### Screenshots:

No query
Screenshot 2024-02-13 102913

Query
Screenshot 2024-02-13 102902

### DIFF
--- a/ui/src/pages/search/authors/index.tsx
+++ b/ui/src/pages/search/authors/index.tsx
@@ -87,7 +87,7 @@ type Props = {
 const Authors: Types.NextPage<Props> = (props): React.ReactElement => {
     const router = Router.useRouter();
     // Query params
-    const [query] = React.useState(props.query ? props.query : '');
+    const { query = '' } = router.query as Interfaces.AuthorSearchQuery;
     const [limit, setLimit] = React.useState(props.limit ? parseInt(props.limit, 10) : 10);
     const [offset, setOffset] = React.useState(props.offset ? parseInt(props.offset, 10) : 0);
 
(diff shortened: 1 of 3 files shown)


ui/src/pages/search/authors/index.tsx
```diff
@@ -87,7 +87,7 @@ type Props = {
 const Authors: Types.NextPage<Props> = (props): React.ReactElement => {
     const router = Router.useRouter();
     // Query params
-    const [query] = React.useState(props.query ? props.query : '');
+    const { query = '' } = router.query as Interfaces.AuthorSearchQuery;
```
adwearing-jisc, line 90: Intriguing. So was this not defaulting to an empty string?

adwearing-jisc: Oh didn't notice you changed props.query to router.query

adwearing-jisc (approved):

ui/src/pages/search/authors/index.tsx
finlay-jisc: Yeah, I think the issue was that the search form submit was only pushing the query term to router. So in order to get the SWR to re fetch we need to take the "query" var from the router instead of props. As it was, it worked after a refresh because getServerSideProps would get search results again based on the query param, but not from "in page" searches.

PR #606, opened 2024-02-15
OC-808: Back end prep for indicating peer review version

finlay-jisc:
The purpose of this PR was to do the back end work involved to store the publication version a link was created to. This will be needed so that we can indicate whether peer reviews were written against the latest version of a publication.

Also renamed some existing fields on links to be clearer about what they actually store/refer to.

Note that the script that will be run as part of deploying this isn't shown in the final file diff; this is because it has to be run while the new "toVersion" field is optional, and this field later becomes mandatory, which would cause the script to fail typescript checking. The script can be found at [this commit](https://github.com/JiscSD/octopus/pull/606/commits/ae5d416339e20207973ab9c3d79047bff51c505f).

The process for deploying this will be:
- Put octopus into maintenance mode
- On the bastion server for the environment being deployed:
    - Check out the [script commit](https://github.com/JiscSD/octopus/pull/606/commits/ae5d416339e20207973ab9c3d79047bff51c505f)
    - Run the available DB migration to add the optional toVersion field and rename the others
    - Run the script to populate toVersion on existing links
    - Check out the up to date branch
    - Run the new available DB migration to make the toVersion field mandatory
- Take octopus out of maintenance mode

---

### Acceptance Criteria:

- The publication version to which links were originally created is now stored in the database
- When a peer review is saved, any links that were to publications which have subsequently had a new version published are updated to refer to that newer published version
- A script is written to populate the “linked to version” on each existing link in the system (to the latest live version)

---

### Checklist:

- [x] Automated tests added
- [ ] Documentation updated

---

### Tests:

API
Screenshot 2024-02-15 105809

E2E
Screenshot 2024-02-14 135518


api/prisma/migrations/20240213151159_link_to_version_pre_script/migration.sql
```diff
@@ -0,0 +1,38 @@
+/*
```
adwearing-jisc, line 1: Not sure what tool you use for migrations, this looks a good candidate to wrap in a transaction if the tool doesn't already

api/prisma/migrations/20240214113648_link_to_version_post_script/migration.sql
```diff
@@ -0,0 +1,8 @@
+/*
+  Warnings:
+
+  - Made the column `versionToId` on table `Links` required. This step will fail if there are existing NULL values in that column.
```
adwearing-jisc, line 4: Take it you can guarantee there are  no NULL values in the column?

api/src/components/publicationVersion/service.ts
```diff
@@ -352,11 +352,50 @@ export const updateStatus = async (id: string, status: I.PublicationStatusEnum)
...
+                where: {
+                    publicationFromId: updatedVersion.versionOf,
+                    draft: true,
+                    versionTo: {
+                        isLatestLiveVersion: false
+                    }
+                }
+            });
+
+            for (const outdatedDraftLink of outdatedDraftLinks) {
```
adwearing-jisc, line 374: Potential to do this in a `promise.all` instead to save some time?

adwearing-jisc (dismissed):

api/prisma/migrations/20240213151159_link_to_version_pre_script/migration.sql
finlay-jisc: These are generated with prisma migrate. From reading about it it doesn't stop you using transactions (or wrap them automatically), you just have to put BEGIN and COMMIT in - I'll add this

api/prisma/migrations/20240214113648_link_to_version_post_script/migration.sql
finlay-jisc: Yes, the script to run in between the migrations will mention if there was a problem filling all of these fields, so we'll know before running this migration

api/src/components/publicationVersion/service.ts
finlay-jisc: In general yes, but since peer reviews can currently only link to one thing, I don't think it's necessary

api/prisma/migrations/20240213151159_link_to_version_pre_script/migration.sql
finlay-jisc: I've now pushed this up - it involved some reverting and cherry picking back of commits because we need a code state to checkout that has only the first migration (but in its transaction-using state) and the script

api/prisma/migrations/20240214113648_link_to_version_post_script/migration.sql
adwearing-jisc: I was messing with prisma migrations myself yesterday afternoon and discovered these warnings are auto generated. Apologies, I thought you wrote this

adwearing-jisc (approved):

PR #607, opened 2024-02-19
OC-768: Create global log in set up script for end to end tests

finlay-jisc:
The purpose of this PR was to speed up our tests by providing a storage file where the tests are being run that individual tests can read and use to authenticate their requests, removing the need to carry out the full log in process in each test that requires being logged in.

Although this PR will show a lot of changes there are basically only a few things being changed, but they affected the tests in lots of places.
- Instead of logging in, tests now use the new getPageAsUser helper.
- A baseUrl has been defined, so we no longer have to use a constant defined in helpers for the base URL. We can use relative paths for navigation instead.
- In publish.e2e.spec.ts, new functions have been made to standardise repeated behaviour and cut code duplication:
    - requestApproval
    - createPublishReadyPublication
    - addCoAuthorsAndRequestApproval
- Removed 'Co Authors appear properly in the Approvals Tracker' test because the only thing it was doing was checking presence of locators that we can check in 'Create a PROBLEM publication with co-authors', which has an identical setup.
- Other than this there are some minor additions to reduce flakiness and cosmetic changes from saving files with prettier.

---

### Acceptance Criteria:

- The end to end tests have a setup script that runs beforehand and logs in each test user
- The tests can use the output of this setup script to skip logging in each time

---

### Checklist:

- [x] Automated tests added
- [x] Documentation updated

---

### Tests:

E2E:
Screenshot 2024-02-19 161401

### DIFF
--- a/e2e/.gitignore
+++ b/e2e/.gitignore
@@ -2,6 +2,7 @@ node_modules/
 /test-results/
 /playwright-report/
 /playwright/.cache/
+/playwright/.auth/
 
 # local env files
 .env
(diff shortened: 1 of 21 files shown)


adwearing-jisc (approved):
Looks good. Haven't had opportunity to use storageState in playwright myself, good excuse to use it here

PR #609, opened 2024-02-20
OC-786: Use newer versions of github actions

finlay-jisc:
The purpose of this PR was to address warnings that are showing up telling us to update our github actions to stop using out of date versions of nodeJS. We should update the following wherever we use them:

- actions/checkout (to v4)
- actions/setup-node (to v4)

---

### Acceptance Criteria:

- We are using v4 of the "checkout" and "setup-node" github actions
- The warnings shown relating to this no longer appear

---

### Checklist:

- [ ] Local manual testing conducted
- [ ] Automated tests added
- [ ] Documentation updated

---

### Tests: N/A

---

### Screenshots:

This is the output of the API and UI tasks run just before this change (note warnings):

Screenshot 2024-02-20 131635
Screenshot 2024-02-20 131652

And on the task run triggered by pushing this update, with no warnings:
Screenshot 2024-02-20 131624
Screenshot 2024-02-20 131645

### DIFF
--- a/.github/workflows/api-build.yml
+++ b/.github/workflows/api-build.yml
@@ -7,10 +7,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
 
       - name: Setup Node 18.x Environment
-        uses: actions/setup-node@v2
+        uses: actions/setup-node@v4
         with:
           node-version: "18.x"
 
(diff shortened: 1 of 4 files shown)


.github/workflows/api-build.yml
```diff
@@ -7,10 +7,10 @@ jobs:
...
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
 
       - name: Setup Node 18.x Environment
-        uses: actions/setup-node@v2
+        uses: actions/setup-node@v4
         with:
           node-version: "18.x"
```
adwearing-jisc, line 15: I know 18.x is still in support, but it's only security support, 20 is the one in active support. Could look to move to that version if it's supported? Having issues over on Azure and terraform with 20 not being fully supported yet, not sure how well AWS/github actions are doing

adwearing-jisc (approved):

.github/workflows/api-build.yml
finlay-jisc: I'd like to use 20 but I've left at 18 for now to match the project, as we still develop + run octopus with node 18 at the moment. It would be good to bump this up to v20 for the whole project soon; I would include the github actions as part of that

PR #611, opened 2024-02-21
OC-784: Automatically zip up pub router lambda code

finlay-jisc:
The purpose of this PR was to use terraform to automatically generate a zip archive of the code for our pub router lambda function. This will mean we don't have to keep the zip in our codebase and regenerate it manually each time it changes.

I tested this by running a new terraform apply on the int environment. At first, it changed the lambda function in place because the file name and source_code_hash changed in the state. I think this will just happen once after changing it from the manually-maintained zip file. After that, I deleted the resulting zip file, ran a terraform plan again, and the zip file was regenerated without suggesting any changes to the terraform state. Also confirmed that the pub router function was working as before.

---

### Acceptance Criteria:

- Terraform generates the zip file on the fly for the pubrouter lambda function.

---

### Checklist:

- [ ] Automated tests added
- [ ] Documentation updated


adwearing-jisc:
> After that, I deleted the resulting zip file, ran a terraform plan again, and the zip file was regenerated without suggesting any changes to the terraform state. Also confirmed that the pub router function was working as before.

Worth doing another test with this. Make a change to your code, then do a new terraform plan and see if it suggests a change. Had issues in the past with terraform not catching changes like this.

finlay-jisc:
> > After that, I deleted the resulting zip file, ran a terraform plan again, and the zip file was regenerated without suggesting any changes to the terraform state. Also confirmed that the pub router function was working as before.
> 
> Worth doing another test with this. Make a change to your code, then do a new terraform plan and see if it suggests a change. Had issues in the past with terraform not catching changes like this.

Good point, I didn't think about testing it with changed code. It looks okay though - I just tried adding a space to a console.log and this brought up a change in the TF plan. If I put it back how it was, it didn't suggest a change.

adwearing-jisc (approved):

PR #633, opened 2024-03-28
OC-324: Implement Content Security Policy (CSP)

finlay-jisc:
The purpose of this PR was to send a Content Security Policy to the Octopus front end application, as an added layer of security that helps to detect and mitigate certain types of attacks, including Cross Site Scripting (XSS) and data injection attacks.

---

### Acceptance Criteria:

- A CSP response header is sent on Octopus and is restrictive as we can make it without breaking parts of the application.

---

### Checklist:

- [ ] Automated tests added
- [ ] Documentation updated

---

### Tests:

E2E:
Screenshot 2024-03-28 115923

### DIFF
--- a/ui/next.config.js
+++ b/ui/next.config.js
@@ -1,4 +1,18 @@
 /** @type {import('next').NextConfig} */
+const CSPDirectives = [
+    "default-src 'self';",
+    "script-src 'self' 'unsafe-eval' 'unsafe-inline';",
+    "style-src 'self' 'unsafe-inline';",
+    `connect-src 'self' *.api.octopus.ac http://127.0.0.1:4003 https://api.octopus.ac https://api.ror.org https://cdn.contentful.com;`,
+    'img-src https:;',
+    'media-src https://octopus-promo-video.s3.eu-west-1.amazonaws.com https://www.youtube.com;',
+    "font-src 'self';",
+    "object-src 'none';",
+    "base-uri 'self';",
+    "form-action 'self';",
+    "frame-ancestors 'none';",
+    'upgrade-insecure-requests;'
+];
 const nextConfig = {
     reactStrictMode: true,
 
@@ -13,6 +27,19 @@ const nextConfig = {
                 source: '/documentation/api',
                 destination: 'https://jiscsd.github.io/octopus',
                 permanent: true
+            }
+        ];
+    },
+    async headers() {
+        return [
+            {
+                source: '/(.*)',
+                headers: [
+                    {
+                        key: 'Content-Security-Policy',
+                        value: CSPDirectives.join(' ')
+                    }
+                ]
             }
         ];
     }
--- a/ui/next.config.js
+++ b/ui/next.config.js
@@ -1,18 +1,48 @@
 /** @type {import('next').NextConfig} */
 const CSPDirectives = [
-    "default-src 'self';",
-    "script-src 'self' 'unsafe-eval' 'unsafe-inline';",
+    "default-src 'none';",
+    `script-src 'self' ${process.env.NODE_ENV === 'development' ? "'unsafe-eval'" : ''};`,
     "style-src 'self' 'unsafe-inline';",
-    `connect-src 'self' *.api.octopus.ac http://127.0.0.1:4003 https://api.octopus.ac https://api.ror.org https://cdn.contentful.com;`,
-    'img-src https:;',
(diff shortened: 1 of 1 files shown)


ui/next.config.js
```diff
@@ -1,4 +1,18 @@
 /** @type {import('next').NextConfig} */
+const CSPDirectives = [
+    "default-src 'self';",
```
adwearing-jisc, line 3: Wouldn't this be better as 'none'? Ideally everything will be set to something or 'self', everything else 'none'.

ui/next.config.js
```diff
@@ -1,4 +1,18 @@
...
+    "script-src 'self' 'unsafe-eval' 'unsafe-inline';",
```
adwearing-jisc, line 4: Why unsafe-eval and unsafe-inline? These should be removed. If you need them locally for hot reload, you can specifically add them for dev mode.

ui/next.config.js
```diff
@@ -1,4 +1,18 @@
...
+    "style-src 'self' 'unsafe-inline';",
```
adwearing-jisc, line 5: As above, unsafe-inline shouldn't be used.

adwearing-jisc: Though I recall having issues with tailwind and needing unsafe-inline (not sure what CSS you're using)

ui/next.config.js
```diff
@@ -1,4 +1,18 @@
...
+    `connect-src 'self' *.api.octopus.ac http://127.0.0.1:4003 https://api.octopus.ac https://api.ror.org https://cdn.contentful.com;`,
+    'img-src https:;',
```
adwearing-jisc, line 7: Should this have a defined list?

ui/next.config.js
```diff
@@ -1,4 +1,18 @@
...
+    'media-src https://octopus-promo-video.s3.eu-west-1.amazonaws.com https://www.youtube.com;',
+    "font-src 'self';",
+    "object-src 'none';",
+    "base-uri 'self';",
+    "form-action 'self';",
+    "frame-ancestors 'none';",
```
adwearing-jisc, line 13: missing frame-src

ui/next.config.js
```diff
@@ -15,6 +29,19 @@ const nextConfig = {
                 permanent: true
             }
         ];
+    },
+    async headers() {
```
adwearing-jisc, line 33: Are you adding other headers beyond CSP? Examples for hedd here - https://github.com/JiscSD/hedd-axolotl/blob/main/axolotl/next.config.js (note frame-src is set to * due to payment page loading iframes from banks)

adwearing-jisc (changes requested):
Questions surrounding using unsafe-inline and unsafe-eval. Ideally these would be removed in CSP rules.

ui/next.config.js
finlay-jisc: Yeah, it makes sense for the catch-all to be less permissive. I've set it to 'none'

ui/next.config.js
finlay-jisc: I based it off of the example on the NextJS docs which also had those - I've now set it to be 'self' only unless local, in which case it also has 'unsafe-eval', which then works OK. Will see how it works on the deployed environment

ui/next.config.js
finlay-jisc: The styles weren't loading unless I used unsafe-inline. We are using tailwind, so maybe that's why

ui/next.config.js
finlay-jisc: I would suggest not because there's only this place that it's being used, but if we needed it elsewhere a dedicated list would make more sense

ui/next.config.js
finlay-jisc: Added this in the latest commit

ui/next.config.js
finlay-jisc: Have included most of the extra headers from the hedd example, only leaving out X-XSS-Protection because MDN said it wasn't needed unless you're really trying to support legacy browsers

ui/next.config.js
adwearing-jisc: :+1: That's fair enough, annoying issue with tailwind. unsafe-inline on styles isn't nearly as bad as scripts. Necessary evil for now.

ui/next.config.js
adwearing-jisc: Great, after I sent you the link I revisited Hedd's and noticed they probably need updating as some are replaced by CSP.

PR #643, opened 2024-04-15
OC-854: Route links on int via correct DOI string

finlay-jisc:
The purpose of this PR was to ensure that DOI links point to the actual DOI address, rather than to Octopus directly, to provide persistence if Octopus were to go down. On int (and locally), these DOI links should be to [https://handle.test.datacite.org/](https://handle.test.datacite.org/) to avoid 404ing.

---

### Acceptance Criteria:

- DOI links on int in the publication metadata box and on PDFs point to DOI URLs beginning with https://handle.test.datacite.org/, following by the DOI ID string
---

### Checklist:

- [ ] Automated tests added
- [ ] Documentation updated

---

### Tests:

UI
Screenshot 2024-04-15 105436

API
Screenshot 2024-04-15 110311

E2E
Screenshot 2024-04-15 105408

---

### Screenshots:

DOI links will appear like this on int.
Screenshot 2024-04-15 112339

### DIFF
--- a/api/src/lib/helpers.ts
+++ b/api/src/lib/helpers.ts
@@ -713,7 +713,7 @@ export const formatAffiliationName = (affiliation: I.MappedOrcidAffiliation): st
     }${organization.address.country}`;
 };
 
-const doiLinkBase = `https://doi.org/`;
+const doiLinkBase = `https://${process.env.STAGE === 'prod' ? 'doi.org' : 'handle.test.datacite.org'}/`;
 
 export const createPublicationHTMLTemplate = (
     publicationVersion: I.PublicationVersion,
--- a/e2e/tests/PageModel.ts
+++ b/e2e/tests/PageModel.ts
@@ -92,7 +92,7 @@ export const PageModel = {
             // coi
             'h2:has-text("Conflict of interest")'
         ],
-        doiLink: 'aside [aria-label="DOI link: https://doi.org/10.82259/cl3fz14dr0001es6i5ji51rq4"]',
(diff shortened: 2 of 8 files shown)


api/src/lib/helpers.ts
```diff
@@ -713,7 +713,7 @@ export const formatAffiliationName = (affiliation: I.MappedOrcidAffiliation): st
     }${organization.address.country}`;
 };
 
-const doiLinkBase = `https://doi.org/`;
+const doiLinkBase = `https://${process.env.STAGE === 'prod' ? 'doi.org' : 'handle.test.datacite.org'}/`;
```
adwearing-jisc, line 716: Possible alternative, put the doi url in environment variables. can just use ${process.env.DOI_URL} for example. Other than code style preferences, benefit is incase that url changes or you need to introduce a different url for int in the future, it's easy to change in env vars.

adwearing-jisc (approved):
Just a possible alternative to retrieve the DOI urls, more down to style preferences though.

api/src/lib/helpers.ts
finlay-jisc: That was my initial plan actually; I thought it would be easier not to have it in the env vars just to keep that list as short as it can be and give new people 1 less thing to look up / understand before they can get started. It's true about the ease of changing though.

adwearing-jisc: Ah interesting thought, maybe I should learn a lesson from this, I just make the env files bigger and bigger then regret it when I have to do a clean clone of the repo.

PR #666, opened 2024-06-25
OC-879: Create new topics for ARI DB mapping

finlay-jisc:
The purpose of this PR was to have a way of creating a set of new topics in bulk so that we are prepared to map topics from the ARI DB into octopus topics.

After this is approved the script will be run on int, the results checked, and run on prod if observed to have worked correctly. The dry run setting will be changed manually on the bastion server where the script is run.

---

### Acceptance Criteria:

- Create new topics based upon mapping sheet
    - Linked to existing parents where details
    - Where no parent exists, repeat through script until parent does exist
    - If possible, return list of created topic URLs/IDs w/ names

---

### Checklist:

- [ ] Automated tests added
- [ ] Documentation updated


api/prisma/createTopics.ts
```diff
@@ -0,0 +1,286 @@
...
+
+        if (
+            item.newParentCreationId2 &&
+            !data.find((potentialParent) => potentialParent.creationId === item.newParentCreationId2)
+        ) {
+            valid = false;
+            messages.push(`Creation ID ${item.creationId} has an invalid new parent creation ID 2`);
+        }
+
+        if (!dryRun) {
```
adwearing-jisc, line 93: Could do this even if not dry run? It isn't updating anything

adwearing-jisc: I'll try this again 😂 

Could potentially do this whether dry run or not. It's only finding things, it isn't updating anything

api/prisma/createTopics.ts
```diff
@@ -0,0 +1,286 @@
...
+    dryRun?: boolean
+): Promise<CreatedTopics> => {
+    let { created, deferred } = await createOrDeferTopics(data, {}, environment, dryRun);
+    console.log(
+        `First run complete; deferred ${deferred.length} topics and created ${Object.keys(created).length} topics.`
+    );
+    const maxRuns = 5;
+    let runCount = 1;
+
+    while (deferred.length && runCount < maxRuns) {
```
adwearing-jisc, line 255: What happens to the parent topics you couldn't create if you hit the max runs?

api/prisma/createTopics.ts
finlay-jisc: Yeah, good point. I think working locally I was conflating "dry run" with "not touching the database" which isn't strictly true. I'll remove this check

api/prisma/createTopics.ts
finlay-jisc: I'll add something to report back on which ones couldn't be created

adwearing-jisc (approved):

adwearing-jisc (approved):

PR #670, opened 2024-07-10
OC-835: Direct publishing via API

finlay-jisc:
The purpose of this PR was to create new functionality that allows organisational accounts to create publications that go straight to LIVE state.

---

### Acceptance Criteria:

- An API key protected endpoint or other function is available to create a draft for a new version or entirely new publication on an organisational account, including options for:
    - Corresponding author (organisation) account ID
    - Title
    - Linked items (Publication ID)
        - Optional. If blank, use the default publication (see OC-824)
        - Topic IDs are only accepted for research problems
    - Main text (HTML)
    - Language (always eng)
    - Keywords
    - Conflict of interest statement declaration (always no)
    - Source (new field, text, always populated with “ARI”)
    - External ID (new field, int, optional unless “Source” is populated)
- This endpoint can trigger publication 

---

### Checklist:

- [x] Automated tests added
- [x] Documentation updated

---

### Tests:

API
Screenshot 2024-07-10 141014

E2E
Screenshot 2024-07-10 151838


api/docs/api.yml
```diff
@@ -1,1511 +1,1579 @@
-openapi: 3.0.0
```
adwearing-jisc, line 1: So.. erm.. what changed here? :joy:

api/src/components/link/service.ts
```diff
@@ -124,3 +126,137 @@ export const canLinkBeCreatedBetweenPublicationTypes = (fromType, toType) => {
...
+            valid: true,
+            toPublication: {
+                publicationId: toPublication.id,
+                versionId: toLatestLiveVersion.id
+            }
+        };
+    } catch (err) {
+        console.log(err);
+
+        return { valid: false, details: { code: 500, message: 'Unknown server error.' } };
```
adwearing-jisc, line 260: This might come back to bite you in the future, could you make this message more helpful?

api/src/components/publication/controller.ts
```diff
@@ -56,45 +58,138 @@ export const getSeedDataPublications = async (
...
+                            message: validateLink.details.message
+                        });
+                    } else {
+                        links.push(validateLink.toPublication);
+                    }
+                }
+            }
+
+            // All proposed topics must exist.
+            if (topicIds?.length) {
```
adwearing-jisc, line 137: Do you need the if check here? Is this checking the array is non empty or that topicIds is not null/undefined? For non empty, the for..of will just not loop and all will be fine.

adwearing-jisc: Can't quite follow what's happening here, can topicIds be null/undefined?

api/docs/api.yml
finlay-jisc: Ooh, the diff was a lot easier to make sense of in VS code! Basically, I changed the query param, request body schema and response for the publication create endpoint to reflect the changes in this PR.

api/src/components/link/service.ts
finlay-jisc: I see what you mean, how about something like: 'Link validation failed because of an unexpected error.'?

api/src/components/publication/controller.ts
finlay-jisc: I don't think there's a need for the "?.length" part here now you mention it, so have just removed those.

api/src/components/link/service.ts
adwearing-jisc: That is better, does the `console.log(err)` go anywhere?

api/src/components/publication/controller.ts
adwearing-jisc: :+1: Ok so something about problems and default topics and publications not having topics. I'll take your word there :) Looks better without the `?.length` bit anyway

api/src/components/link/service.ts
finlay-jisc: It will end up in the lambda logs in cloudwatch, so should be useful to investigate when we see that generic 500 error.

adwearing-jisc: :+1: That's what I was looking for. Good stuff!

adwearing-jisc (dismissed):

api/src/components/link/service.ts
finlay-jisc: Cool, I have pushed the new error message.

adwearing-jisc (approved):

PR #674, opened 2024-07-18
OC-883: Determining how to handle ARIs

finlay-jisc:
The purpose of this PR was to create a function that takes mapped data from an ARI question and either triggers creation or re-version of a publication from it (depending on whether one already exists for the given ARI question ID).

---

### Acceptance Criteria:

- When an ARI is ingested:
    - If the ARI has already been ingested previously and has no changes, no action is taken
    - If the ARI has already been ingested previously and has changes, the corresponding research problem is re-versioned
    - If the ARI has not been ingested previously, a new research problem is created
    - The questionID field is validated to ensure that ARIs are not created erroneously

---

### Checklist:

- [x] Automated tests added
- [ ] Documentation updated

---

### Tests:

API
Screenshot 2024-07-18 104955

E2E
Screenshot 2024-07-18 134404


api/jest-setup.ts
```diff
@@ -0,0 +1,7 @@
+// Ensure that process.env values are defined when jest runs code, to match the behaviour
+// configured by serverless's "useDotenv" option (which also uses dotenv-expand).
+
+import * as dotenv from 'dotenv';
+import { expand } from 'dotenv-expand';
```
adwearing-jisc, line 5: What does expand do?

api/src/lib/helpers.ts
```diff
@@ -355,3 +355,33 @@ export const validateEmail = (email: string): boolean => {
 export const replaceHTMLLineBreaks = (html: string): string => {
     return html.replace(/\n|\r\n|\n\r|\r/g, '<br>');
 };
+
+// Check if two arrays are equal.
+export const compareArrays = <T>(a: Array<T>, b: Array<T>): boolean => {
```
adwearing-jisc, line 360: Is it worth having a unit test to confirm this works as expected?

adwearing-jisc (approved):

api/jest-setup.ts
finlay-jisc: It interprets parts of environment variables as bash parameters and expands them. So something like

`MYVAR=drink${FAVOURITE_DRINK}`

Might have a value of drinktea when the env var is actually used. Serverless framework uses the expand library if you tell it to pick up dotenv config. This was causing problems for me because I have a value with a dollar sign in it that was behaving differently in serverless to jest. In serverless, the dollar sign needed escaping, but in jest it didn't because it wasn't using expand until I told it to. So I added that to this setup file.

api/src/lib/helpers.ts
finlay-jisc: Yeah, good call. I'll write some.

adwearing-jisc (approved):

PR #679, opened 2024-08-21
OC-907: Only use one version for ARIs

finlay-jisc:
The purpose of this PR was to change the way we handle incoming ARIs that correspond to an existing octopus publication, but have changes that need to be imported. An issue was anticipated that big updates in the ARI database could cause a lot of updates to octopus publications, causing them to be reversioned, and lots of DOIs to be minted. To prevent this, and in a break from the ordinary "human" publishing workflow, the existing publication version (of which there will only be 1) will be updated in place rather than reversioned. There will only ever be 1 "publication" DOI and 1 "version" DOI.

---

### Acceptance Criteria:

When an ARI is updated:
- The version 1 of the publication is updated directly instead of a version 2 being created
- No new version DOI is minted 
- PDFs are regenerated
- The publication is re-indexed in search

---

### Checklist:

- [x] Automated tests added
- [ ] Documentation updated

---

### Tests:

API
Screenshot 2024-08-20 165128

E2E
Screenshot 2024-08-21 104447


api/src/lib/integrations/ariUtils.ts
```diff
@@ -241,47 +241,43 @@ export const handleIncomingARI = async (question: I.ARIQuestion): Promise<I.Hand
...
-                    success: false,
-                    message: `Created a new version (id: ${newVersion.id}) of the existing publication for ARI question: ${question.questionId}, but failed to update its status to LIVE.`
-                };
+
+        // If user changed, update coAuthors.
+        if (changes.userId) {
+            // Delete old coAuthor.
+            // There should only be one coAuthor but since this is an array, loop through to make it clean.
+            await Promise.all(
+                existingVersion.coAuthors.map(async (coAuthor) => {
```
adwearing-jisc, line 260: You can just return the promise from `deleteCoAuthor` rather than awaiting that promise and wrapping the callback in it's own promise.

adwearing-jisc:
```
await Promise.all(existingVersion.coAuthors.map(coAuthor => coAuthorService.deleteCoAuthor(coAuthor.id));
```

adwearing-jisc (dismissed):

adwearing-jisc (approved):

PR #684, opened 2024-09-09
OC-886: Notifying departments about ARI activity

finlay-jisc:
The purpose of this PR was to let users give their consent for the owning government departments of ARI publications to contact them, if they have linked their publication to that ARI.

As well as the work for the item this includes a few other things.
- New helpers to standardise the following:
    - user's full name from a user object
    - publication URL
- Remove the "DUMMY" event type that was only included for initial testing
- Change some variable names in the UI that didn't make complete sense
- Tidy up the lib/email.ts file (mainly by using inline types)

---

### Acceptance Criteria:

- When a hypothesis or research problem is published against a research problem with a “Source” field of “ARI”:
    - When clicking publish, the usual publishing confirmation modal includes extra text reading “Your publication is linked to a UK government department's Area of Research Interest (ARI). If you would like your contact details to be shared with the relevant government department so that they can get in touch with you about your work, please check the box below.
    - Below this, a radio button is present with options for “Yes/No” associated text reading “I agree to having my email, name and the details of this publication shared with the government department(s) that published this ARI.”
        - This button is mandatory; the confirm publication button is disabled until this field has been populated 
- If the checkbox is selected, an email is dispatched upon publication to email(s) associated with the ARI(s) that are linked to the publication. (email copy is in private word doc)
---

### Checklist:

- [x] Automated tests added
- [ ] Documentation updated

---

### Tests:

E2E
Screenshot 2024-09-09 150930

API
Screenshot 2024-09-09 152458

UI
Screenshot 2024-09-09 152543

---

### Screenshots:

Modal with radio button and text
Screenshot 2024-09-09 153418

Email
Screenshot 2024-09-09 133942

### DIFF
--- a/api/prisma/migrations/20240904131219_remove_dummy_event_type/migration.sql
+++ b/api/prisma/migrations/20240904131219_remove_dummy_event_type/migration.sql
@@ -1,0 +1,14 @@
+/*
+  Warnings:
+
+  - The values [DUMMY] on the enum `EventType` will be removed. If these variants are still used in the database, this will fail.
+
+*/
+-- AlterEnum
+BEGIN;
+CREATE TYPE "EventType_new" AS ENUM ('REQUEST_CONTROL');
+ALTER TABLE "Event" ALTER COLUMN "type" TYPE "EventType_new" USING ("type"::text::"EventType_new");
+ALTER TYPE "EventType" RENAME TO "EventType_old";
+ALTER TYPE "EventType_new" RENAME TO "EventType";
+DROP TYPE "EventType_old";
+COMMIT;
--- a/api/src/lib/helpers.ts
+++ b/api/src/lib/helpers.ts
@@ -209,7 +209,11 @@ export const getUserFullName = <T extends { firstName: string; lastName: string 
         return 'Anonymous User';
     }
 
-    return user.firstName + (user.lastName ? ' ' + user.lastName : '');
+    if (user.lastName) {
+        return `${user.firstName} ${user.lastName}`;
+    } else {
+        return user.firstName;
+    }
 };
 
 // Parses args passed to an npm script in the following style:
--- a/api/src/lib/helpers.ts
+++ b/api/src/lib/helpers.ts
@@ -202,6 +202,16 @@ export const abbreviateUserName = <T extends { firstName: string; lastName: stri
     return user.firstName;
 };
 
+export const getUserFullName = <T extends { firstName: string; lastName: string | null } | undefined | null>(
+    user: T
+): string => {
+    if (!user) {
+        return 'Anonymous User';
+    }
+
+    return user.firstName + (user.lastName ? ' ' + user.lastName : '');
+};
+
 // Parses args passed to an npm script in the following style:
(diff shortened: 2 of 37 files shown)


api/prisma/migrations/20240904131219_remove_dummy_event_type/migration.sql
```diff
@@ -0,0 +1,14 @@
...
+
+  - The values [DUMMY] on the enum `EventType` will be removed. If these variants are still used in the database, this will fail.
+
+*/
+-- AlterEnum
+BEGIN;
+CREATE TYPE "EventType_new" AS ENUM ('REQUEST_CONTROL');
+ALTER TABLE "Event" ALTER COLUMN "type" TYPE "EventType_new" USING ("type"::text::"EventType_new");
+ALTER TYPE "EventType" RENAME TO "EventType_old";
+ALTER TYPE "EventType_new" RENAME TO "EventType";
```
adwearing-jisc, line 12: After you've renamed EventType_new to EventType do you need to update the column to use this? Or does postgres track the column renaming?

api/src/lib/helpers.ts
```diff
@@ -202,6 +202,16 @@ export const abbreviateUserName = <T extends { firstName: string; lastName: stri
...
 };
 
+export const getUserFullName = <T extends { firstName: string; lastName: string | null } | undefined | null>(
+    user: T
+): string => {
+    if (!user) {
+        return 'Anonymous User';
+    }
+
+    return user.firstName + (user.lastName ? ' ' + user.lastName : '');
```
adwearing-jisc, line 212: I spent way too long looking into a nicer alternative to this. 

Came up with this:
`return [user.firstName, user.lastName].filter(n => !!n).join(' ')`

I don't think that's any better but I spent time figuring this out so thought I'd share it :joy:

adwearing-jisc (approved):
Lot to review, question on the migration, but other than that looks good

api/src/lib/helpers.ts
finlay-jisc: I can see why you looked into it because my line is pretty hard to read for such a simple idea. Perhaps it would be clearest over multiple lines:

`
if (user.lastName) {
    return user.firstName + ' ' + user.lastName;
} else {
    return user.firstName;
}
`

Not sure if that is any better either!

adwearing-jisc: Use interpolated string and then yeah I think it looks a lot better: `if (user.lastName) { return `${user.firstName} ${user.lastName}`; } else { return user.firstName; }`

adwearing-jisc: Ah github stripped the ` bits, you get the gist though

api/prisma/migrations/20240904131219_remove_dummy_event_type/migration.sql
adwearing-jisc: Ah good so it tracks the name change

api/src/lib/helpers.ts
finlay-jisc: Nice. Made this change now.

PR #687, opened 2024-09-12
OC-910: Update weekly ingest to store and check dates

finlay-jisc:
The purpose of this PR was to have a more reliable way of knowing when to stop our incremental ARI ingest job. Sometimes, real changes on the ARI side could be detected as not changed on the octopus side because they wouldn't result in changes to the octopus publication. Thus this item changes the criteria for stopping the job when:
- 5 ARIs in a row are found that don't present changes to what we have in octopus (previous condition)
- the job has encountered an ARI with dateUpdated before the start time of the most recent previous job (new condition)
    - N.B. the data is sorted by dateUpdated

This is achieved by adding an IngestLog model to the database where we can store data on ingest jobs such as their start and end times and check that on later runs.

In addition, this PR makes these changes:
- "PublicationBookmark" model is removed from prisma schema and where referenced in code
    - this is a legacy thing that should have been removed when PublicationBookmarks were migrated to Bookmark a while ago
- Elasticsearch (re)indexing when a publication is publish is reworked to use an "upsert" rather than a create (preceded by a delete if doc is pre-existing in index)
    - logging is added around this because the incremental ARI ingest seemed to hang at this point on its first run

---

### Acceptance Criteria:

- When the weekly ingest runs, the following is recorded:
    - Source (“ARI Database”)
    - Date & time started
    - Date & time ended
- When the weekly ingest is running, it will stop if both 5 records in a row are unchanged from what has already been ingested, and the last of these five records (?) shows a date updated older than the start time of the previous ingest
- In the post-publish tasks for updated publications, the search indexing step is changed to do an upsert instead of a delete + recreate.
    - Logging is added around this step
---

### Checklist:

- [x] Automated tests added
- [ ] Documentation updated

---

### Tests:

E2E
Screenshot 2024-09-12 093940

API
Screenshot 2024-09-11 163239

### DIFF
--- a/api/src/components/publicationVersion/service.ts
+++ b/api/src/components/publicationVersion/service.ts
@@ -1665,11 +1665,7 @@ export const updateCanonicalDOI = async (
 // Actions that run after a version is published (changes status to LIVE).
 // Pulled out to a separate function because things may need to run when something is
 // published immediately (i.e. not going through full drafting workflow) and bypasses the updateStatus function.
-export const postPublishHook = async (
-    publicationVersion: I.PublicationVersion,
-    skipPdfGeneration?: boolean,
-    forceReindex?: boolean
-) => {
+export const postPublishHook = async (publicationVersion: I.PublicationVersion, skipPdfGeneration?: boolean) => {
     try {
         // Ensure links made from a PEER_REVIEW version point to the latest live version of the target publication.
         if (publicationVersion.publication.type === 'PEER_REVIEW') {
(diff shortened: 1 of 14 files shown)


api/src/components/publicationVersion/service.ts
```diff
@@ -1732,12 +1728,9 @@ export const postPublishHook = async (
...
 
         // (Re)index publication in opensearch.
-        if (publicationVersion.versionNumber > 1 || forceReindex) {
-            // Delete old OpenSearch record.
-            await publicationService.deleteOpenSearchRecord(publicationVersion.versionOf);
-        }
-
-        await publicationService.createOpenSearchRecord({
+        // TODO: remove this extra logging if we don't observe ARI imports stalling here for some time.
+        console.log(`Indexing publication ${publicationVersion.versionOf} in opensearch.`);
```
adwearing-jisc, line 1732: Guess you have a ticket to revisit this?

adwearing-jisc (approved):

api/src/components/publicationVersion/service.ts
finlay-jisc: We do now!

PR #689, opened 2024-09-19
OC-918: Fix high severity vulnerability in micromatch

finlay-jisc:
The purpose of this PR was to address a vulnerability alerted to us by Snyk in the micromatch package.

It's included in the copy-webpack-plugin. I spent some time trying to see if we could stop using this plugin but I don't think we can. We need font+image files to be present in the build so that we can use fs.readFileSync on them when generating PDFs, and webpack doesn't seem to be able to do this by itself without the copy plugin (because the assets aren't included as import/require statements for webpack to detect). So it has been solved with a package.json override.

PDF generation still works fine after the update.

---

### Acceptance Criteria:

- micromatch is installed at >= 4.0.8

---

### Checklist:

- [ ] Automated tests added
- [x] Documentation updated

---

### Tests:

E2E
Screenshot 2024-09-19 110911

API
Screenshot 2024-09-19 112008


adwearing-jisc (approved):
Just enjoying a moment reading the github issues for micromatch. People harassing the maintainers to fix it who don't believe it is an issue.

PR #692, opened 2024-09-24
OC-888: Deploying ARI ingests to prod

finlay-jisc:
The purpose of this PR was to provide a way to manually trigger ARI ingests on prod, rather than only scheduled runs, to avoid timing out the operation or creating a large number of new records unnecessarily.

This PR requires a new environment variable containing recipient emails to be defined before deployment.

Also:
- Reworked serverless offline file to access environment variables directly from .env file, as we no longer need to "pretend" to get these from SSM with the serverless-offline-ssm package (because the offline stuff was split out into its own file).
- Removed a helper that was only used in one file; the rest of the time we access env vars directly anyway without this helper.
- Made some changes to the playwright setup because it suddenly stopped being able to find the env file and asset files.

---

### Acceptance Criteria:

- ~~Run the script from "OC-884: Create script to create publications for existing ARIs" on prod to populate existing ARIs~~
    - This was run but asked to rollback, and now things have changed and it isn't ready to run anymore, so this AC is no longer applicable.
- The weekly ingest on int no longer runs automatically
- An API key protected endpoint is present to trigger the operation on prod or int
    - This endpoint does not trigger if if it is already running
- Upon the completion of the operation, an email is dispatched to (redacted email addresses) including:
    - Number of publications checked
    - How long it took
    - Number of new publications created
    - Number of publications ~~re-versioned~~ updated (we update these in place rather than reversion)
    - Whether an unrecognised department was present
    - Whether an unrecognised topic was present

---

### Checklist:

- [x] Automated tests added
- [ ] Documentation updated

---

### Tests:

API
Screenshot 2024-09-25 102046

E2E
Screenshot 2024-09-25 100711

---

### Screenshots:

Email contents (HTML):
Screenshot 2024-09-18 142623

Email contents (text):
Screenshot 2024-09-18 142911


api/.env.example
```diff
@@ -33,4 +33,10 @@ MAIL_SERVER=localhost
 LOCALSTACK_SERVER=http://localhost:4566
 
 QUEUE_URL=http://localhost:4566/000000000000/your-queue-name
-SQS_ENDPOINT=http://localhost:4566
\ No newline at end of file
+SQS_ENDPOINT=http://localhost:4566
+
+LIST_USERS_API_KEY=12345
```
finlay-jisc, line 38: LIST_USERS_API_KEY and SLACK_CHANNEL_EMAIL have been in the app for a while but were missing from the example env file so added them here as well.

api/src/lib/sqs.ts
```diff
@@ -1,24 +1,21 @@
 import AWS_SQS, { SQS } from '@aws-sdk/client-sqs';
-import * as Helpers from './helpers';
 
-const queueUrl = Helpers.checkEnvVariable('QUEUE_URL');
-const endpoint = Helpers.checkEnvVariable('SQS_ENDPOINT');
+const queueUrl = process.env.QUEUE_URL;
```
adwearing-jisc, line 3: Do you still need to check the environment variable exists? What will happen now if the code can't find the variable?

adwearing-jisc:
Just thinking with the moving from offline SSM. Noticed this with working on KSGW, adding the secrets from SSM to the lambda puts the raw secret in the lambda config. I won't put a screenshot here for obvious reasons but something that should be looked at eventually.

adwearing-jisc (approved):

api/src/lib/sqs.ts
finlay-jisc: Instead of throwing an error when it tries to access the variables on lines 4+5, it would go further and throw an error when it can't connect to the queue properly when it creates the client I think. I feel like it is better with the helper now... I'll put it back

adwearing-jisc: Probably is better, it will give you a clearer error to diagnose.

PR #694, opened 2024-09-26
OC-911: ARI ingest dry runs

finlay-jisc:
The purpose of this PR was to add a dry run feature to the ARI integration. In order to test the ARI integration, and to avoid the risk of the ingest timing out if a high number of ARIs need to be updated/created, a dry run feature is needed to check how the ingest will behave before each run.

---

### Acceptance Criteria:

- The endpoint to run the ingest operation on int or prod now contains a dry run parameter
    - Enabling this option will cause the ingest to run without actually making any changes, including:
        - No new ARIs are created, nor any updates to existing ARIs
        - No updates to the log of previous ingests that was created as part of OC-910
        - The email (see OC-888) is still returned upon completion of the operation, including stats for what would’ve been created
---

### Checklist:

- [x] Automated tests added
- [ ] Documentation updated

---

### Tests:

E2E
Screenshot 2024-09-26 114359

API
Screenshot 2024-09-26 115625

---

### Screenshots:

Note that the double space before "results" was fixed in a later commit after these screenshots were taken.

Dry run email (HTML/text):
Screenshot 2024-09-26 103729
Screenshot 2024-09-26 103839

Subsequent real run email (HTML/text):
Screenshot 2024-09-26 103903
Screenshot 2024-09-26 103910

This is what would be output in the email if the dry run was run without any ARIs having been imported previously (we are aware the topics are not recognised):
Screenshot 2024-09-26 093553

### DIFF
--- a/api/serverless-config-default.yml
+++ b/api/serverless-config-default.yml
@@ -493,14 +493,19 @@ functions:
                 method: GET
                 cors: true
     # Integrations
-    incrementalAriIngest:
+    incrementalAriIngestHttp:
         handler: src/components/integration/routes.incrementalAriIngest
         timeout: 900
         events:
-            - schedule:
-                rate: cron(0 5 ? * TUE *) # Every Tuesday at 5 a.m.
-                enabled: ${self:custom.scheduledAriIngestEnabled.${opt:stage}, false}
             - http:
                 path: ${self:custom.versions.v1}/integrations/ari/incremental
                 method: POST
                 cors: true
+    # Commented out - for the time being ARI ingests will just be manually triggered.
+    # incrementalAriIngestScheduled:
+    #     handler: src/components/integration/controller.incrementalAriIngest
+    #     timeout: 900
+    #     events:
+    #         - schedule:
+    #             rate: cron(0 5 ? * TUE *) # Every Tuesday at 5 a.m.
+    #             enabled: ${self:custom.scheduledAriIngestEnabled.${opt:stage}, false}
--- a/api/src/lib/email.ts
+++ b/api/src/lib/email.ts
@@ -941,7 +941,7 @@ export const incrementalAriIngestReport = async (options: {
                   createdCount + updatedCount / 2
               } additional seconds due to datacite API rate limits while creating/updating publications.`
             : '');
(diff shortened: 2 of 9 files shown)


api/serverless-config-default.yml
```diff
@@ -493,14 +493,19 @@ functions:
...
     # Integrations
-    incrementalAriIngest:
+    incrementalAriIngestHttp:
         handler: src/components/integration/routes.incrementalAriIngest
         timeout: 900
         events:
-            - schedule:
-                rate: cron(0 5 ? * TUE *) # Every Tuesday at 5 a.m.
-                enabled: ${self:custom.scheduledAriIngestEnabled.${opt:stage}, false}
             - http:
```
adwearing-jisc, line 500: Does this need a http endpoint? Can this just be invoked via AWS CLI?

api/src/lib/email.ts
```diff
@@ -956,20 +969,18 @@ export const incrementalAriIngestReport = async (options: {
...
-            options.unrecognisedDepartments.length
-                ? 'Unrecognised departments: "' + options.unrecognisedDepartments.join('", "') + '".'
-                : ''
-        }
-        ${
-            options.unrecognisedTopics.length
-                ? 'Unrecognised topics: "' + options.unrecognisedTopics.join('", "') + '".'
-                : ''
-        }`;
+${intro}
```
adwearing-jisc, line 972: Formatting seems to be off here

adwearing-jisc (approved):

api/serverless-config-default.yml
finlay-jisc: The idea was to add an endpoint so that non-devs could trigger it without needing a dev to be involved.

api/src/lib/email.ts
finlay-jisc: I did this on purpose, the default formatting had it indented which reads better, but the problem was that the indentation carried through into the string and into the email output.

adwearing-jisc: :+1: Didn't think of that

api/serverless-config-default.yml
adwearing-jisc: Ah, didn't realise non-devs needed this

PR #697, opened 2024-10-07
OC-928: Remove emails from JSON exports

finlay-jisc:
The purpose of this PR was to solve an issue in which user emails are displayed in JSON exports for publications. This occurs in both the “user” section (corresponding author) and the co-authors section. This field should be omitted from JSON exports.

While working on this I checked where else email addresses are returned in API responses and corrected those as well.

In general the approach for keeping this clear and making things clearer with regard to typing, I separated these out into separate default + private versions of the same function but am open to critique of this approach.

---

### Acceptance Criteria:

Email addresses are not returned from the JSON export function.

---

### Checklist:

- [x] Automated tests added
- [ ] Documentation updated

---

### Tests:

UI
Screenshot 2024-10-07 145808

API
Screenshot 2024-10-07 122036

E2E
Screenshot 2024-10-07 145044

### DIFF
--- a/api/src/components/publication/service.ts
+++ b/api/src/components/publication/service.ts
@@ -15,191 +15,253 @@ export const isIdInUse = async (id: string) => {
     return Boolean(publication);
 };
 
-export const get = async (id: string) => {
-    const publication = await client.prisma.publication.findUnique({
-        where: {
-            id
-        },
+const defaultPublicationInclude = {
+    versions: {
         include: {
-            versions: {
-                include: {
+            user: {
+                select: {
+                    id: true,
+                    orcid: true,
+                    firstName: true,
+                    lastName: true,
+                    createdAt: true,
+                    updatedAt: true,
+                    role: true,
+                    url: true
+                }
+            },
+            publicationStatus: {
+                select: {
+                    status: true,
+                    createdAt: true,
+                    id: true
+                },
+                orderBy: {
+                    createdAt: 'desc'
+                }
+            },
+            funders: {
+                select: {
+                    id: true,
+                    city: true,
+                    country: true,
+                    name: true,
+                    link: true,
+                    ror: true,
+                    grantId: true
+                }
+            },
+            coAuthors: {
+                select: {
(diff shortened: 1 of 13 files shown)


api/src/components/publication/service.ts
```diff
@@ -15,191 +15,253 @@ export const isIdInUse = async (id: string) => {
...
                         }
                     }
                 }
             }
         }
+    }
+} satisfies Prisma.PublicationInclude;
+
+// Ugly, but extend the default include to include private information.
+const privatePublicationInclude = {
```
adwearing-jisc, line 188: Not sure what to say to this one 🫨

adwearing-jisc: Is this just cloning defaultPublicationInclude?

api/src/components/publicationVersion/service.ts
```diff
@@ -116,23 +132,45 @@ export const getById = (id: string) =>
...
-export const get = (publicationId: string, version: string | number) =>
+export const privateGetById = (id: string) =>
+    client.prisma.publicationVersion.findFirst({
+        where: {
+            id
+        },
+        include: privatePublicationVersionInclude
+    });
+
+const getVersionFilterFromStringOrNumber = (versionFilter: string | number) => ({
```
adwearing-jisc, line 143: Do we need to specify the type in the name? Just `getVersionFilter` should do really. Maybe also restrict it to just a string, you're doing impure functions in a way, the first condition is converting versionFilter to a number which is a hidden side effect (if I'm reading it right)

adwearing-jisc (approved):

api/src/components/publication/service.ts
finlay-jisc: This is now written in a friendlier way.

api/src/components/publicationVersion/service.ts
finlay-jisc: Yep, don't see a reason why this shouldn't be simplified - now done.

PR #698, opened 2024-10-07
OC-899-fix: update prisma binary target

finlay-jisc:
The purpose of this PR was to change the binary targets for prisma. It seems that the nodeJS 20 runtime environment for lambda requires a newer target for prisma to generate client code.


adwearing-jisc (approved):
Handy to know

PR #718, opened 2024-11-12
OC-958: Move publish options to sit above title

finlay-jisc:
The purpose of this PR was to encourage users to publish by moving the options to contribute to a more clearly visible place that reflects their role as a primary function of the system.

While working on this, also made some more changes:
- Move baseURL to endpoints config from api config to avoid circular dependency error.
- Remove unused errors config.
- Attempt to standardise button classes a bit by introducing a variant prop.
- Fix a bug where clicking logout would cause a "Cancel rendering route" nextJS error because it attempts to follow the link and do a router.push() at the same time.

---

### Acceptance Criteria:

- The following options are no longer present in the side bar actions area: 
    - “Write a linked <publication type>”
    - “Write a linked review”
    - “Write a linked research problem” 
- Below the publication title, whilst logged in, blue buttons (styled the same as the related publication area buttons) are present. These buttons are aligned in a row if the browser window width allows it, in the following order:
    - “Write a linked <Publication Type>”
    - “Write a linked Research Problem>” 
    - “Write a Peer Review”

---

### Checklist:

- [ ] Automated tests added
- [ ] Documentation updated

---

### Tests:

UI
Screenshot 2024-11-12 143414

E2E
Screenshot 2024-11-12 142457

---

### Screenshots:

Previous
Screenshot 2024-11-12 143723

New
Screenshot 2024-11-12 143645

### DIFF
--- a/ui/src/components/Button/index.tsx
+++ b/ui/src/components/Button/index.tsx
@@ -3,60 +3,62 @@ import React from 'react';
 import * as Components from '@/components';
 
 type CommonProps = {
-    id?: string;
-    title: string;
-    endIcon?: React.ReactElement;
-    startIcon?: React.ReactElement;
-    onClick?: (e: React.MouseEvent) => void;
-    disabled?: boolean;
-    className?: string;
-    childClassName?: string;
-    textSize?: string;
-    padding?: string;
-    children?: React.ReactNode;
     accordionConfig?: {
         contentElementId: string;
         expanded: Boolean;
     };
+    childClassName?: string;
+    children?: React.ReactNode;
+    className?: string;
+    disabled?: boolean;
+    endIcon?: React.ReactElement;
+    id?: string;
+    onClick?: (e: React.MouseEvent) => void;
+    padding?: string;
+    startIcon?: React.ReactElement;
+    textSize?: string;
+    title: string;
+    variant?: 'underlined' | 'block' | 'block-alt';
 };
 
 type ConditionalProps = { href: string; openNew?: boolean } | { href?: never; openNew?: never };
 
 type Props = CommonProps & ConditionalProps;
 
+const blockClasses =
+    'border-2 px-2.5 text-white-50 shadow-sm focus:ring-offset-2 children:border-0 children:text-white-50 ';
+
 const Button: React.FC<Props> = (props): React.ReactElement | null => {
     const parentStyles = React.useMemo(() => {
-        return `
-            group
-            inline-flex
-            items-center
-            outline-0
-            focus:ring-2
-            focus:ring-yellow-400
-            disabled:select-none
-            disabled:opacity-50
-            disabled:hover:cursor-not-allowed
-            ${props.className ?? ''}
-            `;
-    }, [props.className]);
+        let classes =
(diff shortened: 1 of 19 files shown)


ui/src/components/Button/index.tsx
```diff
@@ -3,60 +3,62 @@ import React from 'react';
...
-            focus:ring-yellow-400
-            disabled:select-none
-            disabled:opacity-50
-            disabled:hover:cursor-not-allowed
-            ${props.className ?? ''}
-            `;
-    }, [props.className]);
+        let classes =
+            'disabled:select-none disabled:opacity-50 disabled:hover:cursor-not-allowed inline-flex items-center w-fit ';
+        switch (props.variant) {
```
adwearing-jisc, line 35: Is this to make tailwind work properly? I was going to suggest maybe doing an interpolated string, but I recall tailwind had issues with that

adwearing-jisc: Something to do with having to be a static string at build time

ui/src/components/Nav/index.tsx
```diff
@@ -122,7 +126,7 @@ const Nav: React.FC = (): React.ReactElement => {
                     {
                         label: 'Log out',
                         value: '#log-out',
-                        onClick: () => handleLogOut()
+                        onClick: handleLogOut
```
adwearing-jisc, line 129: Ah I always liked fixing these

ui/src/components/Publication/PublicationPage/HeaderActions/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+type Props = {
+    authorIds: string[];
+    publicationId: string;
+    publicationType: Types.PublicationType;
+};
+
+const HeaderActions: React.FC<Props> = (props) => {
+    const { user } = Stores.useAuthStore();
+
+    return user ? (
```
adwearing-jisc, line 17: Nicer alternative:

```
if (!user) {
    return null;
}

<div className="col-span-8>
...
</div>

**Review by adwearing-jisc — APPROVED** [review]

File: ui/src/components/Button/index.tsx (hunk 1 above)
Comment (finlay-jisc) [reply ui/src/components/Button/index.tsx:35]: String interpolation works; this is just the way I found that I felt kept the formatting nice and the conditions clear. Seems like I can do it with interpolation like this, if this seems better?

```
const classes = `disabled:select-none disabled:opacity-50 disabled:hover:cursor-not-allowed inline-flex items-center w-fit ${
  props.variant === 'block'
    ? blockClasses + 'bg-teal-600'
    : props.variant === 'block-alt'
      ? blockClasses + 'bg-green-600'
      : 'outline-0 focus:ring-2 focus:ring-yellow-400'
} ${props.className ?? ''}`;
```

adwearing-jisc: 😬 Yeah definitely keep it as you already it. Wonder what I was thinking of then, recall coming across issues todo with dynamic strings and tailwind

finlay-jisc: One thing I like about doing it this way is it's easy for line breaks and spaces from the template string to make it into the class string in the HTML and it can be a bit ugly. That happened with the previous way before this change.

PR #722, opened 2024-11-19
OC-966: Save failing on publication edits

finlay-jisc:
The purpose of this PR was to fix an issue where attempts to save publications with a reasonably long content field were failing. This was occurring because our cloudfront distribution sitting in front of the API gateway was rejecting any request with a body over 16KB in size, due to the managed AWS rules. It's difficult to override this in the console, so as part of this solution I have terraformed the cloudfront distribution and the "packaged" web acl you can create with it. Then I added an override for the applicable rule.

The extra provider alias is because the web ACL needs to live in the us-east-1 region.

Working fine on int environment after applying this terraform. Note: before applying to prod, the cloudfront distribution (and not the web acl) needs to be imported into the resource defined in the module in the prod workspace.

---

### Acceptance Criteria:

Publications with a long content value (there is a word doc attached to the item in JIRA for PoC, but the relevant length is 16KB) can be saved.

---

### Screenshots:

Successful save request with 37KB content value on int environment:
Screenshot 2024-11-19 161520

### DIFF
--- a/infra/create-app/main.tf
+++ b/infra/create-app/main.tf
@@ -79,8 +79,11 @@ module "cloudfront" {
   source       = "../modules/cloudfront"
   environment  = local.environment
   project_name = local.project_name
+  # Multiple providers are needed to create resources in different regions.
   providers = {
-    aws                    = aws
+    # Default provider (default region).
+    aws = aws
+    # Us-east-1 provider, required for cloudfront distribution's web ACL.
     aws.us-east-1-provider = aws.us-east-1-provider
   }
 }
--- a/infra/create-app/main.tf
+++ b/infra/create-app/main.tf
@@ -80,7 +80,7 @@ module "cloudfront" {
   environment  = local.environment
   project_name = local.project_name
   providers = {
-    aws           = aws
-    aws.us-east-1 = aws.useast1
+    aws                    = aws
+    aws.us-east-1-provider = aws.us-east-1-provider
   }
 }
--- a/infra/create-app/main.tf
+++ b/infra/create-app/main.tf
@@ -75,3 +75,12 @@ module "sns" {
   project_name = local.project_name
 }
 
+module "cloudfront" {
+  source       = "../modules/cloudfront"
+  environment  = local.environment
+  project_name = local.project_name
+  providers = {
+    aws           = aws
+    aws.us-east-1 = aws.useast1
+  }
+}
--- a/infra/create-app/provider.tf
+++ b/infra/create-app/provider.tf
@@ -4,7 +4,7 @@ provider "aws" {
 }
 
 provider "aws" {
-  alias   = "useast1"
+  alias   = "us-east-1-provider"
   region  = "us-east-1"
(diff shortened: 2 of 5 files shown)


infra/create-app/provider.tf
```diff
@@ -3,6 +3,12 @@ provider "aws" {
   profile = var.profile
 }
 
+provider "aws" {
+  alias   = "useast1"
```
adwearing-jisc, line 7: Think this could be named better. I got confused in the file above thinking you were specifying the region rather than the alias. I can't think of a decent name though..

infra/create-app/main.tf
```diff
@@ -75,3 +75,12 @@ module "sns" {
   project_name = local.project_name
 }
 
+module "cloudfront" {
+  source       = "../modules/cloudfront"
+  environment  = local.environment
+  project_name = local.project_name
+  providers = {
+    aws           = aws
+    aws.us-east-1 = aws.useast1
```
adwearing-jisc, line 84: What happens here then? Noticed you have the provider set on the module aswell

infra/create-app/provider.tf
finlay-jisc: I'm not super happy with it either, I was getting a few errors saying that "the provider the resource was created with has a different name" or some such when I tried to change it before. But I'll see if I can get around that and come up with a better name

infra/create-app/main.tf
finlay-jisc: It seems that you have to define providers at the root level and pass them down to modules, if you want to use a non-default provider in the module. This bit is how the providers are passed to the module (since we need to create stuff in us-east-1 and eu-west-1 in the same module, we need both). Then in the module itself you define the required_providers block to tell it what to expect from the root level.

infra/create-app/provider.tf
finlay-jisc: Now pushed!

infra/modules/cloudfront/main.tf
```diff
@@ -0,0 +1,207 @@
...
+    override_action {
+      none {
+      }
+    }
+    statement {
+      managed_rule_group_statement {
+        name        = "AWSManagedRulesCommonRuleSet"
+        vendor_name = "AWS"
+        version     = null
+        rule_action_override {
```
adwearing-jisc, line 88: Is this the one that removes the size restrictions?

adwearing-jisc: How does it work? Just override it, don't set anything specific and it's disabled?

finlay-jisc: Yep. The default action is to block, so this is overriding it with count, which means it will just "take note" and move on.

adwearing-jisc: Ah, maybe add a comment here explaining what this is doing. Either way I'll trust your wisdom here.

adwearing-jisc (approved):

PR #727, opened 2024-11-21
OC-963 fix: Get serverless v4 to deploy in CICD

finlay-jisc:
The purpose of this PR was to fix obstacles to the API deploy pipeline working with serverless v4.
- The packaging is broken (it doesn't exclude dev dependencies, as it is supposed to) so had to add more exclusions to get the package size down to a deployable level.
- Removed config to do with the serverless dashboard, because although that would be nice to have, it means managing and rotating an access key of its own. Instead, we can forgo the dashboard and use the jisc license key.
- Add ssm:PutParameter permission to the API deployment IAM role that allow serverless to add an SSM parameter to store the deployment bucket name.

---

### Acceptance Criteria:

Back end deploy github action works with serverless v4.

---

### Screenshots:

Screenshot 2024-11-21 121026


adwearing-jisc (approved):

adwearing-jisc:
Maybe a workaround, not sure how well it will work (if at all). Install serverless globally rather than as a dev dependency. Build/test the code etc as normal, but before packaging for serverless, install the dependencies for production only and use the global serverless to package up.

PR #731, opened 2024-12-02
OC-969: Add "Author Types" filter to search

finlay-jisc:
The purpose of this PR was to allow users to restrict the publication search to display individual vs organisational publications, and to allow PM to differentiate between individual and organisational publications in order to monitor octopus's stats.

This works by:
- Indexing a boolean field in opensearch tracking whether the publication is authored by a user of type "ORGANISATION"
- Filter on that when the queryParam is provided to GET /publications

Also done:
- Use beforeAll instead of beforeEach where appropriate to speed up API tests.
- Write API tests for GET /publications
- Fix flaky opensearch seeding test util function
- Announce search result changes to assistive technologies

---

### Acceptance Criteria:

- A new filter is present on the publications search about the publication types, titled “Author Types”
    - Below this, two checkboxes are present, unselected by default, of “Individual” and “Organisational”
        - These checkboxes filter by organisational and individual publication authors
        - If both of these checkboxes are deselected, all results are displayed (same behaviour as publication types filter).

---

### Checklist:

- [x] Automated tests added
- [ ] Documentation updated

---

### Tests:

UI
Screenshot 2024-12-02 111813

API
Screenshot 2024-12-02 112902

E2E
Screenshot 2024-12-02 135243

---

### Screenshots:
Screenshot 2024-12-02 135820


api/src/lib/testUtils.ts
```diff
@@ -84,32 +84,63 @@ export const openSearchSeed = async (): Promise<void> => {
...
+
+    do {
+        const allResults = await client.search.search({ index: 'publications', body: { query: { match_all: {} } } });
+
+        if (allResults.body.hits.total.value > 0) {
+            return;
+        } else {
+            // Wait a second before trying again.
+            await new Promise((resolve) => setTimeout(resolve, 1000));
+            waitSeconds++;
```
adwearing-jisc, line 138: I know you have the timeout set to 1 second, but each loop would inevitably take longer than a second. Maybe a clearer way is to count the number of attempts rather than number of seconds.

adwearing-jisc (approved):

api/src/lib/testUtils.ts
finlay-jisc: Thanks, that's true! Changed it to waitAttempts

PR #737, opened 2024-12-11
OC-924: Allow ECS task to be triggerable on demand

finlay-jisc:
The purpose of this PR was to build on the proof of concept ECS task and allow it to be triggerable on demand. This is because  we want to use ECS to run incremental ARI imports that are triggerable by a non-dev with an API key.

This also involved removing the `services.tf` file because we actually didn't need a service in the first place. That is more for maintaining open-ended tasks while we just want one-off executions for the time being.

Also cherry-picked some UI fixes from a pending PR to get end to end tests to pass.

---

### Acceptance Criteria:

- The proof of concept ECS task can be triggered on demand.

---

### Checklist:

- [ ] Automated tests added
- [ ] Documentation updated

---

### Tests:

API
Screenshot 2024-12-11 161724

E2E
Screenshot 2024-12-11 160840

---

### Screenshots:

This is the ECS task log stream in Cloudwatch after I triggered the task using the lambda function.
Screenshot 2024-12-11 152102


api/src/components/integration/controller.ts
```diff
@@ -48,3 +48,16 @@ export const incrementalAriIngest = async (
         return response.json(500, { message: 'Unknown server error.' });
     }
 };
+
+export const triggerECSTask = async (event: I.APIRequest): Promise<I.JSONResponse> => {
+    const apiKey = event.queryStringParameters?.apiKey;
```
adwearing-jisc, line 53: Can this be done with middleware instead? Keep it consistent with other routes

api/src/lib/ecs.ts
```diff
@@ -0,0 +1,25 @@
...
+            awsvpcConfiguration: {
+                securityGroups: config.securityGroups,
+                subnets: config.subnetIds
+            }
+        },
+        taskDefinition: config.taskDefinitionId
+    };
+    const command = new RunTaskCommand(input);
+    const response = await client.send(command);
+    console.log(response);
```
adwearing-jisc, line 24: Keeping for the PoC?

infra/modules/ecs/tasks.tf
```diff
@@ -45,3 +45,33 @@ resource "aws_ecs_task_definition" "hello-world" {
     }
   ])
 }
+
+resource "aws_security_group" "hello-world-task-sg" {
+  name                   = "${var.project_name}-hello-world-task-sg-${var.environment}"
+  description            = "Security group for hello world ecs service"
```
adwearing-jisc, line 51: Replace service with task?

Is this the same resource in terraform or did you bin off the old one and create this one?

adwearing-jisc (approved):
Very nice way to do this, good job :+1:

adwearing-jisc:
Curious thought, does Octopus have an admin area?

api/src/components/integration/controller.ts
finlay-jisc: Yes, that's probably better - just changed this here and a couple of other places

infra/modules/ecs/tasks.tf
finlay-jisc: Good catch thanks - probably from before I understood the difference between task and service :)

Yeah I just moved this resource from the services.tf file. It was used by the service, and is now used by the lambda; this file made the most sense as a home for it.

finlay-jisc:
> Curious thought, does Octopus have an admin area?

Not as yet though, but can see how this would slot into that if it did exist.

api/src/middleware/authentication.ts
```diff
@@ -5,36 +5,47 @@ import * as response from 'lib/response';
...
-            if (!optional) {
-                // If there's no user account, and authentication is *not* optional, then the request is blocked.
-                if (!user) {
-                    return response.json(401, { message: 'Please enter either a valid apiKey or bearer token.' });
+            if (endpointSpecificKey) {
+                // The function only accepts one specific api key.)
+                if (apiKey !== endpointSpecificKey) {
+                    return response.json(401, { message: "Please provide a valid 'apiKey'." });
+                }
+            } else {
```
adwearing-jisc, line 20: Think you could do without the else block:

```
            if (endpointSpecificKey) {
                // The function only accepts one specific api key.)
                if (apiKey !== endpointSpecificKey) {
                    return response.json(401, { message: "Please provide a valid 'apiKey'." });
                } else {
                    return;
                }
            }

            const bearerToken = request.event.headers.Authorization;
            ...
```

From what I can figure out, you return 401/403 as you go along, then if the user makes it to the end, assign the user to the event object. In the case of endpoint specific key, user is null so you don't need to assign it? Unless you expect null specifically over undefined.

api/src/middleware/authentication.ts
```diff
@@ -5,36 +5,47 @@ import * as response from 'lib/response';
...
+                    user = authorizationService.validateJWT(bearerToken.split(' ')[1]);
                 }
 
-                // If the user hasn't made their name visible in ORCiD, we want to disallow them from doing most actions.
-                if (!user?.firstName && !user?.lastName && requiresName) {
-                    return response.json(403, {
-                        message:
-                            'No name detected. Please ensure your name visibility is set to "Everyone" or "Trusted parties" on your ORCiD account, then re-authorize at /authorization.'
-                    });
+                if (!optional) {
```
adwearing-jisc, line 32: Could move all this optional stuff up a bit?

```
if (optional) {
    return;
}

 const bearerToken = request.event.headers.Authorization;
...
```

Saves doing all the checks on the api key/jwt only to disregard it because auth is optional.

api/src/middleware/authentication.ts
finlay-jisc: Not sure I follow; the Object.assign won't get called if endpointSpecificKey is defined because of this else block. I think your way would have the same effect - either way, if endpointSpecificKey is defined it's going to ignore the rest from `const bearerToken ...` onwards.

api/src/middleware/authentication.ts
finlay-jisc: I want to leave the possibility that an endpoint doesn't require a valid authenticated user, but if there is one, the behaviour differs depending on it or uses their details in some way. I think this is done in some places.

Anyway, the early return can still be put in if there is no key/token so I have done that.

PR #750, opened 2025-01-09
OC-989: API report of publications including versions

finlay-jisc:
The purpose of this PR was to assist in tracking KPIs, and provide an easy way for other systems to ingest a log of Octopus publications by providing an endpoint that will return publication version details in a reporting format.

I thought this made most sense to do by repurposing the GET publication versions endpoint rather than adding a new one.

---

### Acceptance Criteria:

- An endpoint is present to request a list of every publication version ever published, including:
    - Versioned DOI
    - Version publish date
    - Versionless DOI
    - Publication type
- Filters are present to exclude/include organisational and individual authors, and specify date ranges
- Seed data publications are excluded by default

---

### Checklist:

- [x] Automated tests added
- [x] Documentation updated

---

### Tests:

API
Screenshot 2025-01-09 090819

E2E
Screenshot 2025-01-08 163319

### DIFF
--- a/api/src/components/publicationVersion/schema/getAll.ts
+++ b/api/src/components/publicationVersion/schema/getAll.ts
@@ -1,13 +1,13 @@
 import * as Enum from 'enum';
 import * as I from 'interface';
 
-const getAll: I.JSONSchemaType<I.OpenSearchPublicationFilters> = {
+const getAll: I.JSONSchemaType<I.GetPublicationVersionsQueryParams> = {
     type: 'object',
     properties: {
         type: {
             type: 'string',
             pattern: `^((${Enum.publicationTypes.join('|')})(,)?)+$`,
-            default: 'PROBLEM,PROTOCOL,ANALYSIS,REAL_WORLD_APPLICATION,HYPOTHESIS,DATA,INTERPRETATION,PEER_REVIEW'
+            nullable: true
         },
         authorType: {
             type: 'string',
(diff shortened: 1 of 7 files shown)


api/src/components/publicationVersion/schema/getAll.ts
```diff
@@ -1,13 +1,13 @@
...
 import * as I from 'interface';
 
-const getAll: I.JSONSchemaType<I.OpenSearchPublicationFilters> = {
+const getAll: I.JSONSchemaType<I.GetPublicationVersionsQueryParams> = {
     type: 'object',
     properties: {
         type: {
             type: 'string',
             pattern: `^((${Enum.publicationTypes.join('|')})(,)?)+$`,
-            default: 'PROBLEM,PROTOCOL,ANALYSIS,REAL_WORLD_APPLICATION,HYPOTHESIS,DATA,INTERPRETATION,PEER_REVIEW'
```
finlay-jisc, line 10: There is no need for this parameter to pass this default value - it's the same as not passing a value. So I changed it to be optional.

api/src/components/publicationVersion/service.ts
```diff
@@ -248,6 +248,84 @@ export const getAllByPublicationIds = async (ids: string[]) => {
     return mappedResults;
 };
 
+export const getAllForReporting = async (options: I.GetPublicationVersionsReportingOptions) => {
+    const where: Prisma.PublicationVersionWhereInput = {
```
adwearing-jisc, line 252: Same readability issue as the other PR where a SPROC maybe beneficial (though not as bad). Wouldn't let it block you for now, one for the future maybe

adwearing-jisc (approved):

PR #753, opened 2025-01-16
OC-925: Configure ECS task to run on a schedule

finlay-jisc:
The purpose of this PR was to set up the ARI import ECS task to run automatically at a scheduled time. This should finish off the ECS solution, ready for its ongoing usage. On Tuesday mornings, an incremental update runs on int, and at the same time, a dry run incremental update runs on prod. The results will be checked on int and the output report emails will be reviewed before triggering a real incremental update on prod via the API endpoint.

The API code around the trigger endpoint has been tidied up and the naming has been updated to refer to ARI imports specifically rather than generic ECS triggering.

The code hasn't been deployed to prod yet but I have double checked that the dry run override works on the int trigger.

---

### Acceptance Criteria:

- A schedule rule exists that triggers the ARI import function at 5am on int and prod
    - On prod, this execution is a dry run
- ARI import runs can be triggered (dry runs and actual runs) via the API endpoint

---

### Checklist:

- [ ] Automated tests added
- [x] Documentation updated

---

### Tests:

API
Screenshot 2025-01-16 142844

E2E
Screenshot 2025-01-16 141927

### DIFF
--- a/infra/modules/ecs/schedule.tf
+++ b/infra/modules/ecs/schedule.tf
@@ -1,18 +1,4 @@
-locals {
-  only_on_int_mapping = {
-    int  = 1
-    prod = 0
-  }
-  only_on_int = local.only_on_int_mapping[terraform.workspace]
-  only_on_prod_mapping = {
-    int  = 0
-    prod = 1
-  }
-  only_on_prod = local.only_on_prod_mapping[terraform.workspace]
-}
-
 # Code adapted from https://medium.com/@igorkachmaryk/using-terraform-to-setup-aws-eventbridge-scheduler-and-a-scheduled-ecs-task-1208ae077360
-
 resource "aws_iam_role" "scheduler" {
   name = "cron-scheduler-role-${var.environment}-${var.project_name}"
   assume_role_policy = jsonencode({
@@ -65,9 +51,8 @@ resource "aws_iam_policy" "scheduler" {
   })
 }
 
-resource "aws_scheduler_schedule" "int_ari_import_cron" {
-  count = local.only_on_int
-  name  = "ari-import-schedule-int"
+resource "aws_scheduler_schedule" "ari_import_cron" {
+  name = "ari-import-schedule-int"
 
   flexible_time_window {
     mode = "OFF"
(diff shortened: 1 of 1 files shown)


api/src/components/integration/service.ts
```diff
@@ -14,6 +14,19 @@ import * as I from 'interface';
  *       recent successful ingest (if this start time is available).
  */
 export const incrementalAriIngest = async (dryRun: boolean, reportFormat: I.IngestReportFormat): Promise<string> => {
+    // Check if a process is currently running.
+    const lastLog = await ingestLogService.getMostRecentLog('ARI', true);
+
+    if (lastLog && !lastLog.end) {
```
adwearing-jisc, line 20: Would be nice to have some tests for this

infra/modules/ecs/schedule.tf
```diff
@@ -0,0 +1,145 @@
...
+        ],
+        Effect   = "Allow",
+        Resource = [aws_sqs_queue.scheduler-dlq.arn]
+      }
+    ]
+  })
+}
+
+resource "aws_scheduler_schedule" "int_ari_import_cron" {
+  count = local.only_on_int
```
adwearing-jisc, line 69: Can all this only_on_int and only_on_prod be replaced by passing in a stage variable? Then doing the count based on the stage being prod or int/dev. Feels more simplified than the mapping above

infra/modules/ecs/schedule.tf
```diff
@@ -0,0 +1,145 @@
...
+
+  schedule_expression = "cron(0 5 ? * TUE *)" # Run every Tuesday at 5AM
+
+  target {
+    arn      = aws_ecs_cluster.ecs.arn
+    role_arn = aws_iam_role.scheduler.arn
+
+    # Override container command to do a dry run instead of a real one.
+    # The output will be checked before manually triggering a real run using the API.
+    input = jsonencode({
```
adwearing-jisc, line 117: Might be a good idea to set this based on the stage (see previous comment). Just set it to null if on dev/int. Then you don't need separate terraform resources for each environment

adwearing-jisc (approved):
Few potential improvements. No deal breakers, I think it's fine without, just possible to improve readability and terraform usage.

infra/modules/ecs/schedule.tf
finlay-jisc: As per your other comment this count has been removed because both environments are using the same resource.

PR #774, opened 2025-02-27
OC-1011: Avoid excess scrolling when using pagination on author pages

finlay-jisc:
The purpose of this PR was to avoid the scrolling functionality triggered by next/prev buttons taking the user away from the current area of the page. If these next/prev buttons are far down the page, previously they would scroll the view to the top of the page, far above the set of results being browsed. This work limits the scroll to the top of the relevant area.

This also contains some edits to the end to end login helper because that was being flaky, and to the publication search result because slight differences in the formatted date (not meaningful) were causing hydration warnings because it was different between client and server.

---

### Acceptance Criteria:

- Using the pagination options on the author page publications list scrolls the user to the top of the list
- Using the pagination options on the author page red flags list scrolls the user to the top of the list

---

### Checklist:

- [x] Automated tests added
- [ ] Documentation updated

---

### Tests:

UI
Screenshot 2025-02-27 111908

E2E passed but WSL crashed halfway through (I ran the ones it didn't get to afterwards) so I don't have a screenshot but please take my word for it!


ui/src/components/Search/PaginatedResults/index.tsx
```diff
@@ -12,13 +11,24 @@ type Props = {
     noResultsMessage?: string;
     offset: number;
     results: React.ReactNode;
+    scrollFunction?: () => void;
```
karmseveeratjisc, line 14: my React is a little rusty but could not the default function below be simply specified here? then you dont need a conditional

karmseveeratjisc (approved):
one comment but very minor

ui/src/components/Search/PaginatedResults/index.tsx
karmseveeratjisc: sorry, you would put it below:

```
const PaginatedResults: React.FC<Props> = ({ document.getElementById(articleId)?.scrollIntoView({ behavior: 'smooth' }) }, scrollTargetRef?: React.ForwardedRef<HTMLElement>) => {
```
but suspect that might have issues with articeId

karmseveeratjisc:
```
const PaginatedResults: React.FC<Props> = ({ document.getElementById(useId())?.scrollIntoView({ behavior: 'smooth' }) }, scrollTargetRef?: React.ForwardedRef<HTMLElement>) => {
```

might work...

finlay-jisc: From trying it out I would have to use the `useId` hook outside the component to ensure it has the same value for the function and the element's ID, but it has to be used in the component body. I think I'll have to leave it as is

PR #773, opened 2025-02-26
OC-1003: Add maths plugin to text editor

finlay-jisc:
The purpose of this PR was to add the mathematics plugin to the tiptap text editor we use. This will allow users to input mathematical formulae to their publication content.

---

### Acceptance Criteria:

- KaTeX formulae entered into main text are rendered on publication pages and PDFs
- KaTeX formulae entered into main text are hidden from other views such as search result tiles 

---

### Checklist:

- [x] Local manual testing conducted
- [ ] Automated tests added
- [x] Documentation updated

---

### Tests:

UI
Screenshot 2025-02-26 154155

E2E
Screenshot 2025-02-26 140033

---

### Screenshots:

The expression is "raw" while editing it, and highlighted grey when the `$$` delimiters have been set around it
Screenshot 2025-02-26 154017

The expression renders when not being edited
Screenshot 2025-02-26 154028

Formulae in publication body, using the sample from the plugin documentation
Screenshot 2025-02-26 115306

Formulae in PDF
Screenshot 2025-02-26 115324

Formulae removed from search result
Screenshot 2025-02-26 115240

PR #775, opened 2025-03-05
OC-1003 fix: Don't exclude entities + katex from deployment package

finlay-jisc:
The purpose of this PR was to fix an issue with the work in OC-1003 (PR #773). The newly added modules entities and katex were excluded from the serverless deployment as they were not needed outside of PDF generation, which has a separately configured deployment package. However, because they're still imported in helpers, this caused a module not found exception.


karmseveeratjisc (approved):
serverless fun...

PR #778, opened 2025-03-07
OC-1014: New endpoint to fetch user's drafts

finlay-jisc:
The purpose of this PR was to do some back end set up for allowing octopus users to choose a draft they have access to for linking. It specified a new endpoint but it makes more sense to add an option to the existing "get user publications" endpoint to filter to initial drafts only.

The item didn't stipulate this, but non-initial drafts (version 2+) don't apply to this use case, hence the filtering by version number when the new parameter is provided.

It also needs to be able to exclude certain publications, as the user will not need to see the draft they are trying to link from in the search results.

There was version filtering in place already which is a bit confusing, but that was to return publications with at least one version matching the filter. This new filter returns publications with only 1 version which is a bit different, so I thought it made sense to be its own parameter.

---

### Acceptance Criteria:

- A new endpoint fetches all drafts that the user is an author on for use on the “linked publications” section of the draft edit workflow

---

### Checklist:

- [x] Automated tests added
- [ ] Documentation updated

---

### Tests:

API
Screenshot 2025-03-07 091506

E2E
Screenshot 2025-03-06 163504

### DIFF
--- a/api/src/components/user/service.ts
+++ b/api/src/components/user/service.ts
@@ -265,18 +265,18 @@ export const getPublications = async (
         ...authorshipFilter,
         ...versionStatusFilter,
         // If a query is supplied, the query matches the latest live title.
-        ...(params.query
-            ? {
-                  versions: {
-                      some: {
-                          isLatestLiveVersion: true,
-                          title: {
-                              search: Helpers.sanitizeSearchQuery(params.query)
-                          }
-                      }
-                  }
-              }
-            : {})
+        ...(params.query && {
+            versions: {
+                some: {
+                    isLatestLiveVersion: true,
+                    title: {
+                        search: Helpers.sanitizeSearchQuery(params.query)
+                    }
+                }
+            }
+        }),
+        // If a publication is to be excluded, it is not included in the results.
+        ...(params.exclude && { id: { notIn: params.exclude.split(',') } })
     };
 
     const userPublications = await client.prisma.publication.findMany({
--- a/api/src/components/user/service.ts
+++ b/api/src/components/user/service.ts
@@ -239,26 +239,40 @@ export const getPublications = async (
     };
 
     // WHERE clauses related to the requested publication status of publications.
-    const versionStatusFilter: Prisma.PublicationWhereInput =
-        // If the requester is not the account owner, get only publications that have a published version.
-        !isAccountOwner
-            ? { versions: { some: { isLatestLiveVersion: true } } }
-            : // Otherwise...
-            // If only initial drafts are requested, return only these.
-            initialDraftsOnly
-            ? { versions: { every: { currentStatus: 'DRAFT', versionNumber: 1 } } }
-            : // If another version status filter has been supplied, get publications with at least one version with a current status matching the given filters.
-            versionStatusArray
-            ? {
-                  versions: {
-                      some: {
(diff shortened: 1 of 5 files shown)


api/src/components/user/service.ts
```diff
@@ -234,36 +235,48 @@ export const getPublications = async (
...
-            : // But if the user is not the owner, get only publications that have a published version
-              { versions: { some: { isLatestLiveVersion: true } } }),
-        // And, if a query is supplied, where the query matches the latest live title.
-        ...(params.query
+        ]
+    };
+
+    // WHERE clauses related to the requested publication status of publications.
+    const versionStatusFilter: Prisma.PublicationWhereInput =
+        // If the requester is not the account owner, get only publications that have a published version.
```
karmseveeratjisc, line 243: quite hard to read this (as evidenced by the comments needed..), could each conditional be easily separated out?

karmseveeratjisc: I think this flattens out as:

!account owner 
{ versions: { some: { isLatestLiveVersion: true } } }

account owner + initial drafts
 { versions: { every: { currentStatus: 'DRAFT', versionNumber: 1 } } }

account owner + !initial drafts + versionStatusArray
{ versions: { some: { currentStatus: in: versionStatusArray } } }

account owner + !initial drafts + !versionStatusArray
{}

api/src/components/user/service.ts
```diff
@@ -234,36 +235,48 @@ export const getPublications = async (
...
+        !isAccountOwner
+            ? { versions: { some: { isLatestLiveVersion: true } } }
+            : // Otherwise...
+            // If only initial drafts are requested, return only these.
+            initialDraftsOnly
+            ? { versions: { every: { currentStatus: 'DRAFT', versionNumber: 1 } } }
+            : // If another version status filter has been supplied, get publications with at least one version with a current status matching the given filters.
+            versionStatusArray
             ? {
                   versions: {
```
karmseveeratjisc, line 253: seems unnecessary and confusing to be multi line here now, it is just

`{ versions: { some: { currentStatus: in: versionStatusArray } } `}``

api/src/components/user/service.ts
finlay-jisc: I'm sure they can; something like this?

```
    let versionStatusFilter: Prisma.PublicationWhereInput;

    if (!isAccountOwner) {
        versionStatusFilter = {
            versions: {
                some: {
                    isLatestLiveVersion: true
                }
            }
        };
    } else {
        if (initialDraftsOnly) {
            versionStatusFilter = {
                versions: {
                    every: {
                        currentStatus: 'DRAFT',
                        versionNumber: 1
                    }
                }
            };
        } else if (versionStatusArray) {
            versionStatusFilter = {
                versions: {
                    some: {
                        currentStatus: {
                            in: versionStatusArray
                        }
                    }
                }
            };
        } else {
            versionStatusFilter = {};
        }
    }
```

api/src/components/user/service.ts
finlay-jisc: This is laid out differently after the conditional structure was changed above.

karmseveeratjisc: much clearer

karmseveeratjisc (approved):

PR #782, opened 2025-03-17
OC-1019: Peer reviews do not populate DOIs with relevant links

finlay-jisc:
The purpose of this PR was to fix an issue where we were assigning incorrect "relatedIdentifiers" on the payloads we send when creating/updating DOIs. Peer reviews should have a relationship of "Reviews" to their linked publication, but they were set with "Continues". On the other side of the relationship, reviewed publications needed a "IsReviewedBy" relationship to the peer review.

This was harder to write with the code as it was, so I have reworked it quite a bit, moving the code to construct and send different types payloads into its own file and adding test coverage.

---

### Acceptance Criteria:

- Peer reviews include a DOI metadata related item type of “Reviews” linking to the item they review
- When a peer review is published, the DOI metadata for the item it reviews is updated with a “ReviewedBy” related item type linking to the peer review

---

### Checklist:

- [x] Automated tests added
- [ ] Documentation updated

---

### Tests:

API
Screenshot 2025-03-17 144051

E2E
Screenshot 2025-03-17 160047

---

### Screenshots:

From a locally published peer reivew, the following is in its DOI record on datacite test:
Screenshot 2025-03-17 151125

The reviewed publication has this in its DOI record:
Screenshot 2025-03-17 151152

### DIFF
--- a/api/src/components/publicationVersion/service.ts
+++ b/api/src/components/publicationVersion/service.ts
@@ -648,37 +648,55 @@ export const postPublishHook = async (publicationVersion: I.PublicationVersion, 
         // Tasks specific to peer reviews.
         if (publicationVersion.publication.type === 'PEER_REVIEW') {
             // Ensure links made from a PEER_REVIEW version point to the latest live version of the target publication.
-            const outdatedDraftLinks = await client.prisma.links.findMany({
+            const linkedTo = await client.prisma.links.findMany({
                 where: {
                     publicationFromId: publicationVersion.versionOf,
-                    draft: true,
+                    draft: true
+                },
+                select: {
+                    id: true,
+                    publicationToId: true,
+                    publicationTo: {
+                        select: {
+                            doi: true
+                        }
+                    },
                     versionTo: {
-                        isLatestLiveVersion: false
+                        include: defaultPublicationVersionInclude
                     }
                 }
             });
 
-            for (const outdatedDraftLink of outdatedDraftLinks) {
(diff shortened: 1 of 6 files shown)


api/src/components/publicationVersion/service.ts
```diff
@@ -1155,13 +736,25 @@ export const postPublishHook = async (publicationVersion: I.PublicationVersion,
 
         let versionWithDOI: I.PublicationVersion | null = null;
 
-        if (!isExemptFromReversioning(publicationVersion)) {
-            versionWithDOI = await generateNewVersionDOI(publicationVersion, previousVersion);
+        if (!Helpers.isPublicationExemptFromReversioning(publicationVersion.publication)) {
```
finlay-jisc, line 739: No new behaviour here, just being a bit clearer about what is being changed.

api/src/lib/doi.ts
```diff
@@ -0,0 +1,441 @@
+import axios, { AxiosResponse } from 'axios';
```
finlay-jisc, line 1: I've split out DOI payload generation and submission into its own file. The main change was that previously we always sent a payload with all the details of the publication, updating all fields. Datacite's API doesn't require this and the change in the ticket only required updating one field (`relatedIdentifiers`) to show the relationships between publications.

With these changes there are now functions to send payloads to datacite that are agnostic about what data they send (could be lots of fields or just one). There is also a bit more separation of concerns regarding whether we are updating a canonical (entire publication) or version-specific DOI.

PR #788, opened 2025-04-01
OC-1034: Allow retain approval flag to be updated via API

finlay-jisc:
The purpose of this PR was to begin the work to allow coauthors to "retain" their approval, once given to a publication.

These first tasks were to:
- Create a DB migration for the new field
- Accept and save it in the updateCoAuthor endpoint, if it's provided
    - This is the endpoint that will be used to update this field
- Return it in some responses that will need it
- Accept it in the updateAll coauthors endpoint

---

### Acceptance Criteria:

- Retain approval flag can be updated via a new or existing co-author endpoint

---

### Checklist:

- [x] Automated tests added
- [ ] Documentation updated

---

### Tests:

API
Screenshot 2025-04-01 111216

E2E

### DIFF
--- a/api/src/components/coauthor/controller.ts
+++ b/api/src/components/coauthor/controller.ts
@@ -29,25 +29,6 @@ export const getAll = async (
 
         const coAuthors = version.coAuthors;
 
-        const correspondingAuthor = coAuthors.find((coAuthor) => coAuthor.linkedUser === version.createdBy);
-
-        // enforce adding corresponding author if it's missing - this will fix old publications which don't have the corresponding author in the coAuthors list
-        if (!correspondingAuthor) {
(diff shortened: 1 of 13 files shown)


api/src/components/coauthor/controller.ts
```diff
@@ -29,25 +29,6 @@ export const getAll = async (
 
         const coAuthors = version.coAuthors;
 
-        const correspondingAuthor = coAuthors.find((coAuthor) => coAuthor.linkedUser === version.createdBy);
```
finlay-jisc, line 32: A script has been run recently to backport a fix to publications that had no entry for the main author in the coauthors array. We now expect the coauthors array to always include the main author, so I've removed several of these bits of code that are mocking the entry up. We relied on these while the data was in an incorrect state but this is no longer the case.

karmseveeratjisc (approved):

PR #819, opened 2025-06-23
OC-1087: Notification system database fields

danielberigoi:
The purpose of this PR was to add the new fields that we need for the notification system:
1. A new "Notification" model
2. A new "UserSettings" model
3. Update the "User" model to include the notifications
4. Update the "User" model to include the user settings
5. New notification enums

---

### Acceptance Criteria:
- All existing tests are passing and no existing functionality is affected

---

### Checklist:

- [ ] Automated tests added
- [ ] Documentation updated

---

### Tests:

Tests will be added with https://jiscdev.atlassian.net/browse/OC-1088

---

### Screenshots:
Screenshot 2025-06-23 at 12 49 48


finlay-jisc (approved):
Technically this looks good, but I was just wondering about the fields and how it will work with the planned solution.

It looks like the Notification model maps to 1 email (since it has title, content, and so on). But it also maps to one specific entity as well. My understanding of the bulletins was that they would gather up 1 or more "events" (e.g. this publication was re-versioned, another publication was red flagged) and send one email listing details of them all. So I'm a bit confused about why the model seems to have some fields relevant to the email (title, content, status), and also some fields relevant to the event (entityId).

All that said, none of this has been explicitly specified in Jira, and this is your solution, so there's no problem here necessarily. I'm curious about how the process of logging each separate event and bundling them into one bulletin email per user will work. Also this data architecture is not set in stone until we start to store real data in there, so can always be tweaked later.

danielberigoi:
@finlay-jisc I've simplified the notification payload + added an action type field

casianazhr (approved):
LGTM

